PR #10, opened 2023-09-05
feat: use OCI registry as a configuration storage

### LINKED ISSUE #2: Configuration service
Configuration service accepts required image configuration as an input, and produces a hash (ID) of that configuration which will be input to the image service.

Configuration doesn't contain the following:

* Talos version
* Output format

Talos version can be changed (to upgrade/downgrade) with the same configuration, and output format is driven by the image service.

Not every configuration is compatible with every output format.

```[tasklist]
### Tasks
- [x] Configuration Input Format
- [ ] Storage Backend
- [x] Configuration API
```

smira:
Drop additional configuration ID HMAC, and instead take direct sha256 of the contents. This matches content-addressable scheme of the OCI registry, so we can use it directly.

Fixes #2


internal/configuration/storage/registry/registry.go
```diff
@@ -0,0 +1,167 @@
...
+func (w *layerWrapper) Digest() (v1.Hash, error) {
+	return v1.Hash{
+		Algorithm: digest.Canonical.String(),
+		Hex:       w.id,
+	}, nil
+}
+
+// Compressed returns an io.ReadCloser for the compressed layer contents.
+//
+// In fact, it returns raw contents.
```
frezbo, line 141: is this because we just upload the raw config?

smira: it's the interface of the library we use to access the registry, it's a bit more specific to "real" images, but we can still store raw data using it

frezbo (approved):
:cool:

PR #15, opened 2023-09-06
feat: implement registry frontend

### LINKED ISSUE #6: Registry Frontend
The most common type will be downloading `installer` image.

E.g. `docker pull registry.svc/talos/installer-<config>:<talos-version>`

smira:
This allows to pull an installer image for a given version of Talos and configuration.

The actual image is served from the registry, the image service is only a frontend that redirects to the registry.

Fixes #6


internal/integration/registry_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	img, err := descriptor.Image()
+	require.NoError(t, err)
+
+	layers, err := img.Layers()
+	require.NoError(t, err)
+
+	assert.Len(t, layers, 2)
+
+	assertImageContainsFiles(t, img, map[string]struct{}{
```
frezbo, line 67: should we also check for secureboot assets?

frezbo (approved):

internal/integration/registry_test.go
smira: we should, but Image Service doesn't support it yet, so we can't :) the test would fail for secureboot installer, so we'll have to fix it once we remove `t.Skip()` for secureboot installer

PR #22, opened 2023-09-13
feat: implement support for system extensions

### LINKED ISSUE #13: System extension support

smira:
Fixes #13

This builds on top of extensions catalog (see
https://github.com/siderolabs/extensions/pull/225), and existing support for specifying extension in the flavor.

Image Service resolve the list of extensions requested for a specific version of Talos into a list of container images, pulls them, and attaches them to the image request.

Image Service also provides endpoints to get information about available Talos versions, supported extensions for each version, etc.

I also refactored a bit flow around fetching & verifying image to re-use it in other flows, added support for authentication to the registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Official Image Service is available at [https://imager.talos.dev](https://imager
 
 ## HTTP Frontend API
 
-### `POST /flavor`
+### `POST /flavors`
 
 Create a new image flavor.
 
@@ -30,6 +30,10 @@ The request body is a YAML (JSON) encoded flavor description:
 customization:
     extraKernelArgs: # optional
         - vga=791
+    systemExtensions: # optional
+      officialExtensions: # optional
+        - siderolabs/gvisor
+        - siderolabs/amd-ucode
 ```
 
 Output is a JSON-encoded flavor ID:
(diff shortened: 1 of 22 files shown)


README.md
```diff
@@ -30,6 +30,10 @@ The request body is a YAML (JSON) encoded flavor description:
 customization:
     extraKernelArgs: # optional
         - vga=791
+    systemExtensions: # optional
+      officialExtensions: # optional
```
frezbo, line 34: maybe `trustedExtensions` well probably doesn't matter since many extensions are community contributed :shrug:

internal/artifacts/manager.go
```diff
@@ -10,23 +10,33 @@ import (
...
-	logger      *zap.Logger
+	options       Options
+	storagePath   string
+	flavorsPath   string
+	logger        *zap.Logger
+	imageRegistry name.Registry
+	pullers       map[Arch]*remote.Puller
 
-	flavorsSingleFlight singleflight.Group
+	sf singleflight.Group
```
frezbo, line 32: is this from new go?

frezbo (approved):

internal/artifacts/manager.go
smira: it's pretty old: https://pkg.go.dev/golang.org/x/sync/singleflight

README.md
smira: I think `official` is like blessed by us... basically part of the extensions catalog, could be not built by us in theory

internal/artifacts/artifacts.go
```diff
@@ -37,3 +42,10 @@ const (
 
 // FetchTimeout controls overall timeout for fetching artifacts for a release.
 const FetchTimeout = 20 * time.Minute
+
+// Various images.
+const (
+	InstallerImage  = "siderolabs/installer"
+	ImagerImage     = "siderolabs/imager"
+	ExtensionsImage = "siderolabs/extensions"
```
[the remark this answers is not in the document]
smira: thanks, updated

frezbo:
I guess this needs a talos integration test once image service is deployed

smira:
yes, hopefully we can do that with SecureBoot images for example

PR #23, opened 2023-09-14
feat: implement simple UI for the Image Service

### LINKED ISSUE #18: HTMX frontend

smira:
Fixes #18

The intention is to provide educational frontend so that people can get comfortable using the Image Service by building the appropriate links.


Unix4ever (approved):

.kres.yaml
```diff
@@ -113,4 +113,9 @@ spec:
 ---
 kind: auto.CI
 spec:
-    provider: drone
+  provider: drone
```
frezbo, line 116: this can already use github actions

internal/frontend/http/ui.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/siderolabs/image-service/internal/artifacts"
+	"github.com/siderolabs/image-service/pkg/flavor"
+)
+
+//go:embed css/output.css
+var cssFS embed.FS
+
+//go:embed js/*
+var jsFS embed.FS
```
frezbo, line 29: oh this is cool, didn't know we could do this

frezbo (approved):

.kres.yaml
smira: one step at a time, I need to move to GHA, but in a separate PR :)

PR #26, opened 2023-09-15
chore: rename with new nomenclature

andrewrynhard:
Themes the nomenclature to align with Talos Linux


andrewrynhard:
Just a draft currently.

smira:
I :heart: factory & schematic, there was also `model`, not sure where it should map?

smira:
(we're in progress of moving to GHA, so CI is somewhat down)

andrewrynhard:
> I ❤️ factory & schematic, there was also `model`, not sure where it should map?

Yeah, I am still working that out but I have some ideas. I think the idea of "A factory that accepts schematics to build a specific model" is the way I am trying to make this work out.

andrewrynhard:
I feel like a "model" is `schematic` + `version`. An `image` is then `model` + `path`. I don't think the API is structured in a way currently that makes these a simple find and replace in this PR but I think that could be a good mental model we can reflect in the API by restructuring it. Alternatively, we could do something like `GET /model/:schematic/:version/:path` (just calling images a model from this point on).

Just thinking out loud here.

smira:
I also feel that "blueprint" might sound nicer than "schematic". I think URLs should reflect what people get - image factory produces images. Or should it be a model factory??.

smira:
(trying to make Drone pick it up)

smira (approved):

PR #27, opened 2023-09-18
fix: import Talos with initramfs generation fixes

smira:
* fix depmod for non-current kernel versions
* fix early u-code building

Add tests for initramfs contents.

### DIFF
--- a/internal/integration/initramfs_test.go
+++ b/internal/integration/initramfs_test.go
@@ -1,0 +1,203 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+//go:build integration
+
+package integration_test
+
+import (
+	"bufio"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"slices"
+	"strings"
+	"testing"
+
+	"github.com/siderolabs/gen/optional"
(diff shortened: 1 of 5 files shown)


internal/integration/initramfs_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+			return
+		}
+	}
+}
+
+func assertInitramfs(t *testing.T, path string, expected initramfsSpec) {
+	t.Helper()
+
+	f, err := os.Open(path)
```
frezbo, line 58: not closed?

internal/integration/initramfs_test.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+
+			actualEarlyPaths = append(actualEarlyPaths, record.Name)
+		}
+
+		for _, expectedPath := range expected.earlyPaths {
+			assert.Contains(t, actualEarlyPaths, expectedPath)
+		}
+
+		eatPadding(t, in)
```
frezbo, line 90: ahh so there's padding indeed at start of each section?

internal/integration/initramfs_test.go
smira: it's closed at the very bottom of this function

internal/integration/initramfs_test.go
smira: I think `cpio` adds it, even though I can't find a word about it, but kernel supports it, so we do same

frezbo (approved):

PR #35, opened 2023-09-29
fix: allow pulling installer image from insecure registry

smira:
This relies on Talos change to pull the image.

Probably this should be re-implemented to make Image Factory pull the installer image, and then sotre as an OCI layout cached.


frezbo (approved):
let's add a TODO somewhere

smira:
I'll create an issue instead

PR #41, opened 2023-10-18
feat: use OCI layout when passing images to the imager

### LINKED ISSUE #39: pull the installer image, store in the OCI layout and feed to the imager (all pulls in factory)

smira:
Fixes #39

This fixes for all image pulls to be part of the Image Factory itself, Talos `imager` now never pulls on its own.

Installer image (base) is passed as an OCI layout to the `imager`, and same for extension images (for consistency).

Virtual `schematic` extension is passed as `.tar`, as it doesn't make a difference actually.


internal/artifacts/manager.go
```diff
@@ -216,14 +206,53 @@ func (m *Manager) GetOfficialExtensions(ctx context.Context, versionString strin
...
+		}
+	}
+
+	return ociPath, nil
+}
+
+// GetInstallerImageRef returns the installer image reference for the given version.
+// func (m *Manager) GetInstallerImageRef(versionString string) string {
+// 	return m.imageRegistry.Repo(InstallerImage).Tag("v" + versionString).String()
+// }
```
Unix4ever, line 246: Is keeping this commented block intentinal?

Unix4ever (approved):

internal/artifacts/manager.go
smira: nope... thanks, I forgot to GC it

PR #42, opened 2023-10-23
feat: sign generated installer image

### LINKED ISSUE #29: Sign & verify cached assets

smira:
Image Factory now signs the generated asset using cosign flow with a fixed key. Image Factory also verifies the signature before redirecting to the image. This way we ensure the consistency of the cache.

The signing ECDSA private key (PEM-encoded) should be supplied as `--signing-key-path` flag.

Fixes #29


frezbo (approved):

smira:
added an API to retrieve public signing key from the Image Factory

PR #46, opened 2023-10-26
feat: implement boot asset cache

### LINKED ISSUE #4: OCI Caching Registry

smira:
Fixes #4

This caches the result of the boot asset build (a call to Talos `imager`) so that we don't rebuild the asset twice.

OCI Registry is used as a cache. An internal registry can be used for caching, we don't need to expose it to the world.


internal/asset/cache.go
```diff
@@ -0,0 +1,146 @@
...
+		return nil, fmt.Errorf("failed to get cache image layer size: %w", err)
+	}
+
+	return &remoteAsset{
+		layer: layer,
+		size:  size,
+	}, nil
+}
+
+// Get returns the boot asset from the cache.
```
frezbo, line 104: comment doesn't match

frezbo (approved):

internal/asset/cache.go
smira: and linter is silent :confused:

PR #48, opened 2023-10-30
fix: implement insecure option for cache repository

smira:
I forgot to implement it while doing asset caching.

### DIFF
--- a/cmd/image-factory/cmd/service.go
+++ b/cmd/image-factory/cmd/service.go
@@ -71,7 +71,8 @@ func RunFactory(ctx context.Context, logger *zap.Logger, opts Options) error {
 		return fmt.Errorf("failed to parse self URL: %w", err)
 	}
 
-	repoOpts := []name.Option{}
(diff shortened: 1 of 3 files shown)


frezbo (approved):

cmd/image-factory/cmd/service.go
```diff
@@ -183,9 +183,14 @@ func buildAssetBuilder(logger *zap.Logger, artifactsManager *artifacts.Manager,
 
 	builderOptions.RemoteOptions = append(builderOptions.RemoteOptions, remoteOptions()...)
 
+	repoOpts := []name.Option{}
```
DmitriyMV, line 186: Linter should complain about that no?

Should be
`var repoOpts []name.Option`

DmitriyMV (approved):

cmd/image-factory/cmd/service.go
smira: I don't think it really matters here, but I will fix it

PR #53, opened 2023-11-02
release(v0.1.1): prepare release

smira:
This is the official v0.1.1 release.


frezbo (approved):

frezbo:
maybe rekres also and check if release is fixed?

smira:
> maybe rekres also and check if release is fixed?

good point, will do

PR #58, opened 2023-11-10
feat: implement SecureBoot asset generation

### LINKED ISSUE #19: Signing UKI with KMS key

smira:
Fixes #19

Using Talos implementation of custom SecureBoot signers, provide full implementation of SecureBoot assets signed either by static local PKI or Azure Key Vault reference.


internal/frontend/http/secureboot.go
```diff
@@ -0,0 +1,25 @@
...
+package http
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// handleSecureBootSigningCert handles SecureBoot siging cert PEM.
```
frezbo, line 14:
```suggestion
// handleSecureBootSigningCert handles SecureBoot signing cert PEM.
```

smira: I want them to sing :)

internal/profile/profile.go
```diff
@@ -305,18 +327,25 @@ func EnhanceFromSchematic(ctx context.Context, prof profile.Profile, schematic *
 	}
 
 	// skip customizations for profile kinds which don't support it
-	if prof.Output.Kind != profile.OutKindInitramfs && prof.Output.Kind != profile.OutKindKernel && prof.Output.Kind != profile.OutKindInstaller {
+	//
+	// initramfs/kernel can't carry extra kernel args & META
+	// !secureboot (non-UKI) installer can't carry extra kernel args & META
+	// UKI installer has kernel args embedded in the UKI image
```
frezbo, line 333: if we want to support, we could later add as efi addons and sd-boot would pick them up

PR #62, opened 2023-11-24
Update readme to highlight registration is required

Ornias1993:
It's not fully clear that creation through the API is required for images to be build.
This change highlights that for images to be made available an API call is mandatory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ Output is a JSON-encoded schematic ID:
 {"id":"2a63b6e7dab90ec9d44f213339b9545bd39c6499b22a14cf575c1ca4b6e39ff8"}
 ```
 
-This ID can be used to download images with this schematic.
+After creation ID can be used to download images with this schematic.
+Creation through the factory is required for the images associated with the ID to be build.
 
 Well-known schematic IDs:
 
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ boot
 
 The Talos Linux `installer` image is used for the initial install and upgrades.
 It can be pulled from the Image Factory OCI registry.
-If the image hasn't been created yet, it will be built on demand automatically.
+If the image hasn't been created yet, it will be built on demand automatically. As long as the ID has been created before.
 
 ### `docker pull <registry>/installer[-secureboot]/<schematic>:<version>`
 


smira:
I don't see the need for these changes. The only way to get the schematic ID is to call the `POST /schematics` API, there is no other documented way to get it. So as long as the API user has a schematic ID, the only way they could get is via the `POST /schematics` API which is idempotent.

So these changes make it more confusing than it was before.

Ornias1993:
You don't write docs because they are clear *to you*, you write them so other developers can use your stuff without making mistakes.

Two (out of: not many) developers making mistakes reading them, is a **very** clear sign your assumption is completely wrong and your defensive attitude without clear arguments besides "I understand it just fine" is, quite frankly, becoming asinine at best.

andrewrynhard:
@Ornias1993 sorry for the confusion on how to use image factory. We certainly want docs to be improved and things clarified. 

@smira maybe we can help get this PR to a place where it is mergable? Where would you suggest we look to help make the docs more clear in this particular case?

smira:
@andrewrynhard I posted my thoughts on this issue above:

> The only way to get the schematic ID is to call the POST /schematics API, there is no other documented way to get it. So as long as the API user has a schematic ID, the only way they could get is via the POST /schematics API.

The @Ornias1993 kept saying that docs are unclear as some schematic ID is not accepted, while it turned out it was never registered, which raises a question - how the ID was acquired in the first place? So I'd guess this ID was acquired by not following the existing documentation (e.g. trying to use code in this repository), but it is not documented as part of the API.

smira:
@andrewrynhard please feel free to re-open/merge this PR, but I personally don't see that any changes are required.

andrewrynhard:
I want to figure out how  we can all win here. We don't want to add confusion to the docs, that is for sure, but there is something missing obviously.

@Ornias1993 do you recall the thought process and/or where you get that ID from?

Ornias1993:
> I want to figure out how we can all win here. We don't want to add confusion to the docs, that is for sure, but there is something missing obviously.
> 
> @Ornias1993 do you recall the thought process and/or where you get that ID from?

As stated in the Issue previously, it was generated by talhelper.
Afaik he generated using the imagefactory packages or something...
Which is already going to be solved downstream, so isn't really the point...

I've had multiple associates read this readme, and no one read it as "registration is a strict requirement"...
It just needs to clearly state that registration of an ID with the factory is strictly required.

There are a lot of usecases where users will hit this, for example: moving to another factory.

PR #63, opened 2023-11-28
fix: properly handle from ghcr.io

smira:
It returns 404 only when trying to pull a layer, but not when we get the manifest.

Refactor all cases when we try to handle registry errors to use a common function.


frezbo (approved):

smira:
Disabled govulncheck here using https://github.com/siderolabs/kres/pull/330

smira:
Once the issue is resolved with `govulncheck`, we can re-enable.

PR #66, opened 2023-11-29
feat: implement HTTP API client

Unix4ever:
The client provides methods to create schematics, Talos versions, and extensions version.

Fixes: #21

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,0 +1,162 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Package client implements image factory HTTP API client.
+package client
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/siderolabs/gen/xerrors"
+
+	"github.com/siderolabs/image-factory/pkg/schematic"
+)
+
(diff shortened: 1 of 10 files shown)


.kres.yaml
```diff
@@ -12,6 +12,10 @@ spec:
       GOOS: linux
       GOARCH: arm64
 ---
+kind: golang.GoVulnCheck
```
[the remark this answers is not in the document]
Unix4ever: yeah, bad merge. Fixed

pkg/client/client.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Client is the Image Factory HTTP API client.
+type Client struct {
+	address string
+	client  http.Client
+}
+
+// New creates a new Omni API client.
+func New(address string, timeout time.Duration) *Client {
```
smira, line 52: it feels it's not `address`, but rather base URL, and we could parse it here to make sure it's valid, and use `.Join` method to append patch to it?

pkg/client/client.go
```diff
@@ -0,0 +1,163 @@
...
+	c := &Client{
+		address: address,
+		client: http.Client{
+			Timeout: timeout,
```
smira, line 56: do we actually need this `Timeout` here? should it be set on the context for each request?

We could also optionally accept `http.Client` from the outside for deep configuration options

Unix4ever: can introduce options list here, use default client if not set and something like `WithClient()` to pass it

pkg/client/client.go
Unix4ever: I do join path in each method. Should be enough I guess

Unix4ever: ok, updated

PR #70, opened 2023-12-15
fix: generation of SBC images

### LINKED ISSUE #65: Raspberry Pi image through factory not working
Test env:
3x raspberry Pi 4, 3 different high grade sd sd cards, official power supply.

Official image:
Loads fine.

After manually generating a raspberry pi url through the image factory with just 2 kernel args:
https://factory.talos.dev/image/8e3aaa276e2b5ccbe7243708ef6a9c9ca80901b986232474eb8259627fec45db/v1.5.5/rpi-arm64.raw.xz

Kernel args where to disable mac-based ethernet interface names and disable ipv6.
Nothing major.

I loaded this to the SD cards the same way as the offical image and it fails booting with 4 blinks.

smira:
The fix is in https://github.com/siderolabs/talos/pull/8076

Also bump the hash to flush the SBC cached images.

Fixes #65


internal/integration/download_test.go
```diff
@@ -233,7 +233,7 @@ func testDownloadFrontend(ctx context.Context, t *testing.T, baseURL string) {
 				t.Run("rpi image", func(t *testing.T) {
 					t.Parallel()
 
-					downloadAssetAndMatchSize(ctx, t, baseURL, emptySchematicID, talosVersion, "metal-rpi_generic-arm64.raw.xz", "application/x-xz", 66574172)
+					downloadAssetAndMatchSize(ctx, t, baseURL, emptySchematicID, talosVersion, "metal-rpi_generic-arm64.raw.xz", "application/x-xz", 107183936)
```
smira, line 236: the image getting correct size is the sign of the fix working :)

internal/profile/hash.go
```diff
@@ -31,6 +31,13 @@ func Hash(p profile.Profile) (string, error) {
 		return "", fmt.Errorf("failed to marshal profile: %w", err)
 	}
 
+	// update the hash value to force rebuild assets once the bug is fixed
+	//
+	// 1. errata https://github.com/siderolabs/image-factory/issues/65
+	if p.Board != "" {
```
frezbo, line 37: oooohhh, I like this way of adding tests for errara

frezbo (approved):

Issue #20, opened 2023-09-13
Support Talos 1.2-1.3


smira:

smira:
Retroactively:

* signed all Talos release container images 1.2.0-1.3.7
* created extension catalog 1.2.0-1.3.7
* signed extension images 1.2.0-1.3.7

PR #76, opened 2023-12-20
fix: update Talos version listing

### LINKED ISSUE #75: :bug: `/version/:version/extensions/official` returns 500 for pre-release Talos versions

smira:
Now Image Factory filters out pre-release versions for all releases but the last one.

In the UI, now pre-release versions are shown.

Return proper 404 not found when someone requests something for an unsupported version.

Fixes #75 

(the versions before 1.4.0 are not yet enabled in Image Factory, see #20)

### DIFF
--- a/internal/artifacts/artifacts.go
+++ b/internal/artifacts/artifacts.go
@@ -56,3 +56,6 @@ const (
 )
 
 const tmpSuffix = "-tmp"
+
+// ErrNotFoundTag tags the errors when the artifact is not found.
+type ErrNotFoundTag = struct{}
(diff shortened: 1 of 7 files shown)


smira:
image

frezbo (approved):
a test maybe?

smira:
> a test maybe?

yep, at least we can try 404...

Unix4ever (commented):
As an option you can also update the client to report a separate error type for 404.
Not a big deal though.

smira:
> As an option you can also update the client to report a separate error type for 404. Not a big deal though.

yep, that also makes sense. we should probably report HTTP errors properly there

smira:
updated with some tests and HTTP error wrapping

PR #78, opened 2023-12-22
feat: implement compatibility with Talos 1.2-1.3

### LINKED ISSUE #20: Support Talos 1.2-1.3

smira:
Lower min version to 1.2.0, pull in fixes for `imager` from Talos.

Add new tests for v1.3.7 as an example.

Fixes #20


internal/integration/registry_test.go
```diff
@@ -69,7 +69,9 @@ func testInstallerImage(ctx context.Context, t *testing.T, registry name.Registr
 	layers, err := img.Layers()
 	require.NoError(t, err)
 
-	assert.Len(t, layers, 2)
+	if talosVersion != "v1.3.7" {
+		assert.Len(t, layers, 2)
```
frezbo, line 73: let's also add some context message for the error, `3 != 2` never really helps :sweat_smile:

frezbo (approved):

internal/integration/registry_test.go
smira: which error? I just skip the assertion for v1.3.7, which doesn't have 2 layers

frezbo: i meant for the `assert` if it fails just says `x !=2` without much context why it failed

smira: I'm not sure how it can be made more helpful? if you go back to the line you know why it failed, but if I make custom message, iirc it will skip the actual `3 != 2`, so you are left to wonder what was the length

smira: oh, I'm wrong, it still prints the assertion

smira: thanks, updated

PR #81, opened 2023-12-26
feat: fetch extensions descriptions from the extensions image

Unix4ever:
Get `descriptions.yaml` from the same image as we use for getting `image-digests`.
Read it and add this data to each extension info.
Return that in the `version/:version/extensions/official` API response.


smira (approved):
should we update the UI then to show them as `label`s or something like that? (for the checkboxes?)

PR #94, opened 2024-03-15
feat: support overlay

frezbo:
Support handling overlays.

Fixes: https://github.com/siderolabs/talos/issues/8355


frezbo:
image

README.md
```diff
@@ -39,6 +39,11 @@ customization:
       officialExtensions: # optional
         - siderolabs/gvisor
         - siderolabs/amd-ucode
+overlay: # optional
+  name: rpi_generic # overlay name
+  image: ghcr.io/siderolabs/sbc-raspberry-pi # overlay image
```
smira, line 44: I wonder if we should swap these, as `image` defines an overlay, and `name` is one of the overlays in the image?

internal/artifacts/schematic.go
```diff
@@ -103,11 +108,24 @@ func schematicExtension(schematicID string) (io.Reader, error) {
...
 		Name:     filepath.Join("rootfs/usr/local/share/schematic", schematicID), // empty file
 		Typeflag: tar.TypeReg,
-		Mode:     0o755,
+		Mode:     0o644,
 	}); err != nil {
 		return nil, fmt.Errorf("failed to write rootfs header: %w", err)
 	}
 
+	if err = tw.WriteHeader(&tar.Header{
+		Name:     "rootfs/usr/local/share/schematic/schematic.yaml",
```
smira, line 117: this would be hard to fetch back from Omni. we can keep it this way as well, but I think we should rather embed the schematic into some field of the `extensions.Metadata` above, so that we can get a schematic back by doing `talosctl get extensions`

internal/asset/asset.go
```diff
@@ -249,7 +250,7 @@ func (b *Builder) build(ctx context.Context, prof profile.Profile, versionString
 		}
 	}
 
-	if prof.Arch == string(artifacts.ArchArm64) {
+	if prof.Arch == string(artifacts.ArchArm64) && !quirks.New(versionString).SupportsOverlay() {
```
smira, line 253: :+1:

pkg/schematic/schematic.go
```diff
@@ -46,6 +48,13 @@ type SystemExtensions struct {
 	OfficialExtensions []string `yaml:"officialExtensions,omitempty"`
 }
 
+// Overlay represents the overlay options for image generation.
+type Overlay struct {
+	Options map[string]any `yaml:"options,omitempty"`
```
smira, line 53: I think we should reverse the order here: `image`, `name` and `options`

smira (approved):

README.md
frezbo: didn't get this, `name` is the overlays in the image, not sure I follow

internal/artifacts/schematic.go
frezbo: I was thinking omni could read back the file?

frezbo: should we change the signature of `extensions.Metadata`?

pkg/schematic/schematic.go
frezbo: that was to make govet happy :sweat_smile:

README.md
smira: I mean order of the fields

internal/artifacts/schematic.go
smira: Omni can't read files in maintenance mode

pkg/schematic/schematic.go
smira: Nolint it!

internal/artifacts/schematic.go
frezbo: hmm maybe i can add another virtual extensions with the name `overlay` and set the `version` field as `<image>:<name>` or something like that

frezbo: hmm, that doesn't work since, we need `options` too

frezbo: > Omni can't read files in maintenance mode

didn't we land support for 1.7 to allow over maintenance? and since overlays are for 1.7+ this should work I believe

smira: no, this doesn't work. I wonder if we could re-use the description or any other free-form field (or add another one) just to put there the schematic as string as YAML

smira (approved):

PR #95, opened 2024-04-04
feat: add cert issuer regexp option

mattiascockburn:
This enables the verification of images if they were signed using different (OIDC) issuers.


smira (changes requested):

smira (approved):
:+1:

PR #96, opened 2024-04-04
feat: add support for -insecure-schematic-service-repository flag

mattiascockburn:
Allow usage of an insecure schematics repository


smira (approved):
:+1:

PR #107, opened 2024-04-17
fix: generation of overlay installer images

smira:
Pull in fixes from Talos, rekres, add a test for overlay installer generation.

Due to the nature of the fixes, the first fully working installer should come with the next release of Talos Linux.


internal/profile/hash.go
```diff
@@ -41,6 +41,10 @@ func Hash(p profile.Profile) (string, error) {
 	if p.Output.Kind == profile.OutKindInstaller {
 		hasher.Write([]byte("installer fix #8107"))
 	}
+	// 3. overlay installer layout issues
```
frezbo, line 44: link the github issue also?

frezbo (approved):

internal/profile/hash.go
smira: yep, will do

PR #110, opened 2024-04-29
feat(images): allow key-based image verification as option

mattiascockburn:
This PR enables the image-factory to optionally verify images using a supplied public key. If both OIDC and public key are configured both options will be tried when verifying an image.

### DIFF
--- a/cmd/image-factory/cmd/service.go
+++ b/cmd/image-factory/cmd/service.go
@@ -25,7 +25,9 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sigstore/cosign/v2/cmd/cosign/cli/fulcio"
 	"github.com/sigstore/cosign/v2/pkg/cosign"
+	"github.com/sigstore/cosign/v2/pkg/signature"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	sigstoresignature "github.com/sigstore/sigstore/pkg/signature"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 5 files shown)


cmd/image-factory/flags.go
```diff
@@ -23,6 +23,8 @@ func initFlags() cmd.Options {
 	flag.StringVar(&opts.ContainerSignatureSubjectRegExp, "container-signature-subject-regexp", cmd.DefaultOptions.ContainerSignatureSubjectRegExp, "container signature subject regexp")
 	flag.StringVar(&opts.ContainerSignatureIssuerRegExp, "container-signature-issuer-regexp", cmd.DefaultOptions.ContainerSignatureIssuerRegExp, "container signature issuer regexp")
 	flag.StringVar(&opts.ContainerSignatureIssuer, "container-signature-issuer", cmd.DefaultOptions.ContainerSignatureIssuer, "container signature issuer")
+	flag.StringVar(&opts.ContainerSignaturePublicKeyFile, "container-signature-pubkey", cmd.DefaultOptions.ContainerSignaturePublicKeyFile, "container signature public key")
```
smira, line 26: should say that it's optional

cmd/image-factory/cmd/service.go
```diff
@@ -313,3 +339,48 @@ func loadPrivateKey(keyPath string) (crypto.PrivateKey, error) {
...
+	}
+
+	if !ralgo.Available() {
+		return crypto.SHA256, fmt.Errorf("hash %q is not available on this platform", algo)
+	}
+	return ralgo, nil
+}
+
+func getPublicKeyVerifier(opts Options) (sigstoresignature.Verifier, error) {
+	var keyVerifier sigstoresignature.Verifier
```
smira, line 369: this code and many others won't pass `make lint`

cmd/image-factory/cmd/service.go
```diff
@@ -178,6 +181,8 @@ func runMetricsServer(ctx context.Context, logger *zap.Logger, eg *errgroup.Grou
 }
 
 func buildArtifactsManager(ctx context.Context, logger *zap.Logger, opts Options) (*artifacts.Manager, error) {
+	var err error
+	var artifactsManager *artifacts.Manager
```
smira, line 185: not sure why this declaration is even needed

internal/artifacts/fetch.go
```diff
@@ -113,12 +114,18 @@ func (m *Manager) fetchImageByDigest(digestRef name.Digest, architecture Arch, i
 	// verify the image signature, we only accept properly signed images
 	logger.Debug("verifying image signature")
 
-	_, bundleVerified, err := cosign.VerifyImageSignatures(ctx, digestRef, &m.options.ImageVerifyOptions)
-	if err != nil {
+	var verifyFailedCount int = 0
```
smira, line 117: this might better be extracted to a separate sub-function which verifies, and verification should go until the first successful verification returning the result

internal/artifacts/fetch.go
```diff
@@ -113,12 +114,18 @@ func (m *Manager) fetchImageByDigest(digestRef name.Digest, architecture Arch, i
...
+		if err != nil {
+			verifyFailedCount += 1
+		}
+	}
+	if len(m.options.ImageVerifyOptions) == verifyFailedCount {
 		return fmt.Errorf("failed to verify image signature for %s: %w", digestRef.Name(), err)
 	}
 
-	logger.Info("image signature verified", zap.Bool("bundle_verified", bundleVerified))
+	logger.Info("image signature verified")
```
smira, line 128: this should log how it was verified, so dropping bundle should be reverted

smira (changes requested):

internal/artifacts/fetch.go
mattiascockburn: I now log the method used.

internal/artifacts/fetch.go
mattiascockburn: Done

cmd/image-factory/cmd/service.go
mattiascockburn: Removed the declaration

cmd/image-factory/cmd/service.go
mattiascockburn: Fixed up the code so that make lint passes.

cmd/image-factory/flags.go
mattiascockburn: Done

cmd/image-factory/cmd/service.go
```diff
@@ -313,3 +340,54 @@ func loadPrivateKey(keyPath string) (crypto.PrivateKey, error) {
...
+	return ralgo, nil
+}
+
+func getPublicKeyVerifier(opts Options) (sigstoresignature.Verifier, error) {
+	var keyVerifier sigstoresignature.Verifier
+
+	if len(strings.TrimSpace(opts.ContainerSignaturePublicKeyFile)) > 0 {
+		hashAlgo, _ := getHashAlgo(opts.ContainerSignaturePublicKeyHashAlgo) //nolint:errcheck
+
+		kf, err := os.OpenFile(opts.ContainerSignaturePublicKeyFile, os.O_RDONLY, os.ModePerm)
```
smira, line 376: `kf` is not closed, but easier to replace all of that with `os.ReadFile`

internal/artifacts/fetch.go
```diff
@@ -113,12 +114,21 @@ func (m *Manager) fetchImageByDigest(digestRef name.Digest, architecture Arch, i
...
-	logger.Info("image signature verified", zap.Bool("bundle_verified", bundleVerified))
+	// Determine verification method and log it
+	var vm string
+	if co.SigVerifier != nil {
+		vm = "public key"
+	} else {
+		vm = "certificate subject"
+	}
+
+	logger.Info("image signature verified", zap.String("verification_method", vm))
```
smira, line 131: can we keep the `bundle_verified` and return it from `verifyImageSignatures`?

cmd/image-factory/cmd/service.go
mattiascockburn: Done

internal/artifacts/fetch.go
mattiascockburn: Hope i got it right this time :-)

smira (approved):
I did some cleanups and fixes, the coding style and features were not  consistent with Image Factory source code

PR #122, opened 2024-06-06
feat: allow a user to skip the Talos version filter

mattiascockburn:
This commit introduces a new flag '-skip-version-filter' which skips the filtering of available versions in the artifactManager.


steverfrancis:
What's the use case for this?

mattiascockburn:
@steverfrancis Cheers, we are using a custom versioning scheme based on the Talos version to better distinguish our build from the base Talos release. This way we are able to use these versions as well as development build. It might be better to make this more configurable but simply deactivating the filter is fine for us right now.

PR #132, opened 2024-07-18
chore: make metatadata pkg public

frezbo:
Make metadata pkg public.


smira (approved):
it imports a bit of internal pkgs, might be nice to add something like

```
type Arch = artifacts.Arch
```

to prevent it being internal type

pkg/client/client.go
```diff
@@ -57,6 +57,11 @@ func New(baseURL string, options ...Option) (*Client, error) {
 	return c, nil
 }
 
+// BaseURL returns the base URL of the client.
+func (c *Client) BaseURL() string {
```
frezbo, line 61: this is to do less hacks in TF provider

frezbo:
> it imports a bit of internal pkgs, might be nice to add something like
> 
> ```
> type Arch = artifacts.Arch
> ```
> 
> to prevent it being internal type

good now?

smira (approved):

PR #151, opened 2024-09-12
[WIP] add support for returning just the schematic id without json

TMaYaD:
Client can request to skip the json encoding by requesting `text/plain` in `Accept` header.

This is a non-working PoC written in browser. If you like the direction, I can add tests/clean-up etc and submit a proper PR.


smira:
Thank you, I think this looks a possible solution

PR #187, opened 2025-01-16
feat: update for Talos 1.10 current

smira:
Handle non-signed UKI build, rekres.


utkuozdemir (approved):

dsseng (approved):

.kres.yaml
```diff
@@ -28,8 +28,6 @@ spec:
   customCommands:
     - apk add --no-cache --update
         bash
-        binutils-aarch64
```
smira, line 31: now we generate UKIs native in Go without `objcopy`

frezbo, line 31: do we need a new errata, but since signed uki's are never reproducable, I guess it doesn't add any value

DmitriyMV (approved):

.kres.yaml
smira: hopefully not, they are functionally same (and probably they weren't byte-by-byte same due to external dependency)

PR #189, opened 2025-01-17
feat: refactor platform metadata

smira:
Updates the wizard to be more correct, handle SecureBoot for !metal choice, etc.

Fixes [#10108](https://github.com/siderolabs/talos/issues/10108)


smira:
Metadata now comes from https://github.com/siderolabs/talos/pull/10173

frezbo (approved):

PR #191, opened 2025-01-19
add imgfree to standard.ipxe boot script

lukemilby:
fix for `Initramfs unpacking failed: invalid magic at start of compressed`

While using image generated from the image factory I was getting `Initramfs unpacking failed: invalid magic at start of compressed` when booting. https://github.com/ipxe/ipxe/discussions/1329. I tried the suggestion adding `imgfree` before kernel command and was able to boot into talos.

### DIFF
--- a/internal/frontend/http/secureboot.ipxe
+++ b/internal/frontend/http/secureboot.ipxe
@@ -1,4 +1,5 @@
 #!ipxe
 
+imgfree
 kernel {{ .UKIURL }} {{ .Cmdline }}
 boot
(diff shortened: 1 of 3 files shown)


frezbo:
the image factory ipxe script doesn't load any image before the lines adding `imgfree`, as per docs https://ipxe.org/cmd/imgfree it cleans up previous artifacts, but this is the first one downloaded, so not sure if this is the right fix

frezbo:
> We could probably do `imgfree` still, if Image Factory script is chained from another script? @frezbo ?

oh that makes sense if it's chained

smira (approved):

PR #205, opened 2025-03-05
feat: pull in new Talos machinery

frezbo:
Pull in new Talos machinery, drop dependency on alpine and add required tools from pkgs.


go.mod
```diff
@@ -9,42 +9,42 @@ require (
...
 	github.com/julienschmidt/httprouter v1.3.0
-	github.com/klauspost/compress v1.17.11
+	github.com/klauspost/compress v1.18.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/prometheus/client_golang v1.20.5
 	github.com/siderolabs/gen v0.8.0
-	github.com/siderolabs/go-debug v0.4.0
+	github.com/siderolabs/go-debug v0.5.0
 	github.com/siderolabs/go-pointer v1.0.0
-	github.com/siderolabs/talos v1.10.0-alpha.1.0.20250203100439-e2aa7c98cceb
```
smira, line 18: we have alpha.2 now we could use

smira (approved):

PR #208, opened 2025-03-07
chore: add more tests for talos 1.10

frezbo:
Add more tests for Talos 1.10.
Reduce layers of image-factory container image.
Also drop integration test for non-pull-requests.


Makefile
```diff
@@ -229,6 +229,12 @@ lint: lint-golangci-lint lint-gofumpt lint-govulncheck lint-markdown  ## Run all
 image-image-factory: tailwind  ## Builds image for image-factory.
 	@$(MAKE) registry-$@ IMAGE_NAME="image-factory"
 
+.PHONY: imager-base
+imager-base:
+
+.PHONY: imager-tools
+imager-tools:
```
dsseng, line 236: Wondering whether these should really be empty targets here

frezbo: yeh, side effect of kres :sweat_smile:

smira (approved):

PR #210, opened 2025-03-18
feat: support platform specific installers

frezbo:
Support platform specific installers.

Fixes: https://github.com/siderolabs/talos/issues/10486


internal/frontend/http/ui.go
```diff
@@ -467,7 +467,7 @@ func (f *Frontend) wizardFinal(ctx context.Context, params WizardParams) (string
 			ImageBaseURL: f.options.ExternalURL.JoinPath("image", schematicID, version),
 			PXEBaseURL:   f.options.ExternalPXEURL.JoinPath("pxe", schematicID, version),
 
-			InstallerImage:           fmt.Sprintf("%s/installer/%s:%s", f.options.ExternalURL.Host, schematicID, version),
+			InstallerImage:           fmt.Sprintf("%s/%s-installer/%s:%s", f.options.ExternalURL.Host, params.Platform, schematicID, version),
```
smira, line 470: 1. I think we should only prepend Platform for Talos 1.10+
2. we should update SecureBootInstallImage same way

internal/frontend/http/ui.go
```diff
@@ -636,148 +636,6 @@ func (f *Frontend) handleUIVersionDoc(_ context.Context, w http.ResponseWriter,
 	return getTemplates().ExecuteTemplate(w, "version-doc.html", version)
 }
 
-// handleUIVersions handles '/ui/schematic-config'.
-func (f *Frontend) handleUISchematicConfig(ctx context.Context, w http.ResponseWriter, r *http.Request, _ httprouter.Params) error { //nolint:unused // to be removed
```
smira, line 640: nice catch

internal/integration/download_test.go
```diff
@@ -237,11 +285,60 @@ func testDownloadFrontend(ctx context.Context, t *testing.T, baseURL string) {
...
+					})
+				})
+
 				t.Run("installer image", func(t *testing.T) {
 					t.Parallel()
 
 					downloadAssetAndMatchSize(ctx, t, baseURL, emptySchematicID, talosVersion, "installer-amd64.tar", "application/x-tar", sizePicker(talosVersion, "1.5", 167482880, "1.8", 185155584, "1.9", 136*MiB, "1.10", 127*MiB))
 					downloadAssetAndMatchSize(ctx, t, baseURL, emptySchematicID, talosVersion, "installer-arm64.tar", "application/x-tar", sizePicker(talosVersion, "1.5", 222793728, "1.8", 170119168, "1.9", 126*MiB, "1.10", 116*MiB))
+
+					if quirks.New(talosVersion).SupportsUnifiedInstaller() {
```
smira, line 308: it _might_ be easier to flip this other way around, and do something like this:

```go
t.Run("legacy installer image", func(t *testing.T) {
    if quirks.New(talosVersion).SupportsUnifiedInstaller() {
         t.Skip()
    }

    downloadAssetAndMatchSize...
}

t.Run("uki installer image", func(t *testing.T) {
    if !quirks.New(talosVersion).SupportsUnifiedInstaller() {
         t.Skip()
    }

    downloadInstallerAndValidateUKI...
}
```

internal/frontend/http/ui.go
frezbo: > I think we should only prepend Platform for Talos 1.10+

does it matter, since factory will still handle old `installer` and `installer-secureboot`

smira: > does it matter, since factory will still handle old installer and installer-secureboot

I think we shouldn't confuse users with a sudden change which wasn't there

frezbo:
> we also need to update
> 
> https://github.com/siderolabs/image-factory/blob/e1d19dfbff074ccbec9751ed3c9a3eff8914bf47/internal/profile/profile.go#L433-L439
> 
> otherwise it should ignore extra kernel args for the installers with UKIs
> 
> and we need to add an integration test for it (generating installer image from a schematic with extra kernel args)

I didn't quite get this logic, seems we set extra kernel args for uki which seems correct :thinking:

smira:
> I didn't quite get this logic, seems we set extra kernel args for uki which seems correct 🤔

The problem is that it ignores ExtraKernelArgs for all installers except for SecureBoot ones, while it should actually use them for any installers >= 1.10 (in addition to SecureBoot).

README.md
```diff
@@ -83,7 +83,8 @@ Supported image paths:
 * `initramfs-<arch>.xz` (e.g. `initramfs-amd64.xz`) - initramfs image (including system extensions if configured)
 * `<platform>-<arch>[-secureboot].iso` (e.g. `metal-amd64.iso`) - ISO image
 * `<platform>-<arch>[-secureboot]-uki.efi` (e.g. `metal-amd64-secureboot-uki.efi`) UEFI UKI image (Secure Boot compatible)
-* `installer-<arch>[-secureboot].tar` (e.g. `installer-amd64.tar`) is a custom Talos Linux installer image (including system extensions if configured)
+* `installer-<arch>[-secureboot].tar` (e.g. `installer-amd64.tar`) is a custom Talos Linux installer image for `metal` platform (including system extensions if configured)
```
smira, line 86: we should also update for the registry pulls around line 180

smira (approved):

PR #214, opened 2025-04-04
chore: bump talos machinery

frezbo:
Bump talos machinery.


frezbo:
making sure factory works with xfs quirks

smira (approved):

PR #219, opened 2025-04-22
chore: bump deps

frezbo:
Upgrade to latest Talos beta.


smira (approved):
we also have `PKGS` in `.kres.yaml`, let's bump it to 1.10.0

Issue #218, opened 2025-04-22
CI: nightly run with Talos `main`


smira:
Add a CI cron job which will update Talos in the Image Factory to `main` version and run an integration test.

PR #222, opened 2025-04-29
chore(ci): add an cron ci for talos main integration test

frezbo:
Add a daily scheduled cron for running image-factory from Talos main.

Fixes: #218


Makefile
```diff
@@ -145,7 +145,7 @@ else
 GO_LDFLAGS += -s
 endif
 
-all: unit-tests image-factory image-image-factory imager-base imager-tools integration.test integration tailwind lint
+all: unit-tests image-factory image-image-factory imager-base imager-tools integration.test integration update-to-talos-main integration-talos-main tailwind lint
```
smira, line 148: this is a nit, but probably we shouldn't have this in `all` (and some others as well?)

smira (approved):

Makefile
frezbo: i'll do a kres fix

PR #223, opened 2025-04-29
fix(Dockerfile): fix build on ARM64 Apple Silicon

mattiascockburn:
This enables the build of image-factory on ARM64 systems

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,7 @@ FROM ${PKGS_PREFIX}/zlib:${PKGS} AS pkg-zlib
 FROM ${PKGS_PREFIX}/zstd:${PKGS} AS pkg-zstd
 
 # Installs tailwindcss
-FROM docker.io/oven/bun:1.2.4-alpine AS tailwind-base
+FROM --platform=${BUILDPLATFORM} docker.io/oven/bun:1.2.4-alpine AS tailwind-base
 WORKDIR /src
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -78,7 +78,7 @@ FROM ${PKGS_PREFIX}/zlib:${PKGS} AS pkg-zlib
 FROM ${PKGS_PREFIX}/zstd:${PKGS} AS pkg-zstd
 
 # Installs tailwindcss
-FROM docker.io/oven/bun:1.2.4-alpine AS tailwind-base
+FROM --platform=${BUILDPLATFORM} docker.io/oven/bun:1.2.4-alpine AS tailwind-base
```
smira, line 81: thank you for your PR, but unfortunately this file is automatically generated via `kres`, so we need either a fix to 	`kres` or to `.kres.yaml`

mattiascockburn: Good to know! I'll look into that when i got time.

frezbo: https://github.com/siderolabs/image-factory/pull/222/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R81 should fix this

smira: @mattiascockburn thanks for your PR, I will close it as #222 contains the fix via `.kres.yaml` update

Issue #232, opened 2025-05-15
Some OVA files cannot be imported into vSphere Content Library


olavst-spk:
Hi!

This issue is similar to https://github.com/siderolabs/talos/issues/8148, but I am creating a new one since the old thread has been locked. 

I'm running Talos on VMWare and having some issues with the OVA files from the image factory.  I get the following error when trying to import an OVA of Talos 1.10.1 into the vSphere Content Library:
```
A general system error occurred: The import of library item 623033d0-c198-4ef3-84e8-ed5731007b3a has failed. Reason: Error transferring file disk.vmdk to ds:///vmfs/volumes/65afc448-6f117e7c-4ad9-b8830367db1c//contentlib-c87ae30e-7a02-4c97-8360-2cf68ee1fae5/623033d0-c198-4ef3-84e8-ed5731007b3a/disk_8aa84709-0e7a-4ca9-8fcc-bfce2c08c244.vmdk?serverId=e090c5fa-5586-4112-94af-4eb39f5fcc5f. Reason: IO error during transfer of ds:///vmfs/volumes/65afc448-6f117e7c-4ad9-b8830367db1c//contentlib-c87ae30e-7a02-4c97-8360-2cf68ee1fae5/623033d0-c198-4ef3-84e8-ed5731007b3a/disk_8aa84709-0e7a-4ca9-8fcc-bfce2c08c244.vmdk?serverId=e090c5fa-5586-4112-94af-4eb39f5fcc5f: Pipe closed. 
```
Here is a link to the problematic OVA file: https://factory.talos.dev/image/903b2da78f99adef03cbbd4df6714563823f63218508800751560d3bc3557e40/v1.10.1/vmware-amd64.ova. Other versions (for example 1.10.0) can be imported without any problems.

I did some digging and the problem seems to be that the hash in the manifest file (`disk.mf`) inside the OVA is incorrect. We can see this if we run the OVA through ovftool:
```sh
$ ovftool vmware-amd64.ova foo.ova
Opening OVA source: vmware-amd64.ova
The manifest validates
Opening OVA target: foo.ova
Writing OVA package: foo.ova
Transfer Failed                       
Error: SHA digest of file disk.vmdk does not match manifest
Completed with errors
```

To verify this I used the `--skipManifestCheck` flag to generate a "repaired" OVA file:
```sh
$ ovftool --skipManifestCheck vmware-amd64.ova vmware-amd64-repaired.ova
Opening OVA source: vmware-amd64.ova
[...]

PR #236, opened 2025-05-21
fix: vmware ova generation

frezbo:
Fixes: #232


smira (approved):

smira:
Do we have an integration test with OVA?

frezbo:
> Do we have an integration test with OVA?

nope, adding one

PR #237, opened 2025-05-27
fix: pull in overrides from the overlay profile

smira:
Previously it was completely ignored in the Image Factory.

See https://github.com/siderolabs/talos/issues/11093

Still a question for me what _can_ be overridden, and what is the best way to merge profiles.


internal/profile/profile.go
```diff
@@ -436,6 +440,30 @@ func EnhanceFromSchematic(
...
+			overlayProfileData, err := os.ReadFile(filepath.Join(overlayProfilePath, overlayRef.Name+".yaml"))
+			if err != nil {
+				return prof, fmt.Errorf("error reading overlay profile %s: %w", overlayProfilePath, err)
+			}
+
+			if err := yaml.Unmarshal(overlayProfileData, &overlayProfile); err != nil {
+				return prof, fmt.Errorf("error unmarshalling overlay profile %s: %w", overlayProfilePath, err)
+			}
+
+			if overlayProfile.Output.ImageOptions != nil {
```
smira, line 459: that is the worst part - I basically just copy the bootloader option here, but I don't know what can be taken and what not, or whether we should instead flip the stuff around and start with the initial base profile and apply Image Factory options on top of it?

smira: reworked this part completely

internal/profile/hash.go
```diff
@@ -72,6 +72,11 @@ func Hash(p profile.Profile) (string, error) {
 		hasher.Write([]byte("installer tarball fix #10278"))
 	}
 
+	// 8. Images with overlays do not import properly profile overrides.
```
frezbo, line 75: Let's also add a comment to issue https://github.com/siderolabs/talos/issues/11093

internal/profile/profile.go
```diff
@@ -479,6 +488,54 @@ func EnhanceFromSchematic(
...
+	}
+
+	// zero out fields in the overlay profile which are set explicitly by the Image Factory
+	overlayProfile.Input = profile.Input{}
+	overlayProfile.Arch = ""
+	overlayProfile.Platform = ""
+	overlayProfile.Overlay = nil
+	overlayProfile.Version = ""
+	overlayProfile.SecureBoot = nil
+	overlayProfile.Output.Kind = profile.OutputKind(0)
```
frezbo, line 523: I think calling `outkind.<type>.String()` is more readable

smira: it's not a string here, but rather enum (int).

 I want a zero value here so that `merge.Merge` will skip it

internal/profile/hash.go
smira: thanks, fixed it up

PR #267, opened 2025-07-22
feat: update to Talos 1.11.0-beta.0

smira:
Update Talos, PKGS, rekres and update dependencies.


frezbo (approved):

.kres.yaml
```diff
@@ -186,10 +184,6 @@ spec:
               from: pkg-grub-arm64
               src: /usr/lib/grub
               dst: /usr/lib/grub
-          - copy:
```
frezbo, line 189: don't we need for old versions of installer?

PR #268, opened 2025-07-23
iPXE example is not working due to architecture naming mismatch and wrong factory URL

manumeter:
Here's a (not very nice) example on how to fix these issues. If there's a nicer solution, I'm open for it - that's just how I solved it for now.


smira:
Thanks for spotting this, nice catch. I think we could fix it in the Image Factory itself by normalizing the architecture name. I will take a look

PR #308, opened 2025-09-18
fix: log mount error

irqoff:
When you build Talos 1.10 on 5.15:
```
error:error generating asset: failed to install: failed to install bootloader: error mounting partitions: error mounting /dev/loop3p3: 2 error(s) occurred:
        invalid argument
        timeout
```
Added log about possible differences between the kernel it runs on and the one it is being built on.


smira:
This is not a fix in any way, this would produce a broken image.

PR #317, opened 2025-09-30
feat: support proxying external installer registry

### LINKED ISSUE #313: support reverse proxying to the installer registry
Instead of doing HTTP redirects, supports additionally a mode where IF reverse proxies to the registry

smira:
This basically drops the need for external installer registy endpoint completely: we can now simply have the internal one which will be used by the Image Factory to push images to, and also it will proxy pass access to it.

Fixes #313


smira:
Missing integration tests

Issue #311, opened 2025-09-29
Add support for koanf


shanduur:
Currently, configuration is managed only through flags. This makes it harder to support more complex or layered configuration, and less convenient for users who want to manage settings via files or environment variables.

* Integrate [koanf](https://github.com/knadh/koanf) in addition to flags, to allow:
* Configuration via (one of) YAML/JSON/TOML file.
* Overriding with environment variables.
* Retaining existing flag support for compatibility.

PR #323, opened 2025-10-15
feat: upload aliases from config file

Alex619829:
Adds a flag for specifying the path to a YAML file with extension name aliases.


smira:
let's wait for #311 and circle back to this issue

PR #330, opened 2025-11-05
feat: bump deps

frezbo:
Bump deps.


smira (approved):
if we bump Talos machinery, we should bump `PKGS` in `.kres.yaml` to match Talos 1.12 as well

PR #337, opened 2025-12-10
fix: support new cosign bundle format

smira:
Refactor the image signature verification code, add more tests.

Support both old and new bundle format from cosign.

Update cosign to address
https://github.com/sigstore/timestamp-authority/security/advisories/GHSA-4qg8-fj49-pxjh.


frezbo (approved):
🆒

PR #341, opened 2025-12-17
feat: refactor configuration of image factory

### LINKED ISSUE #311: Add support for koanf

shanduur:
Refactor configuration of Image Factory.
This PR removes most of the flags, and instead allows configuration using config files and/or environment variables.

```console
Usage of image-factory:
      --config configs    Configuration source(s). Can be specified multiple times or as a comma-separated list.
                          Supported forms:
                            env=[PREFIX]        Load configuration from environment variables (optional prefix).
                            FILE                Load configuration from a file; format is inferred from extension.
                            file=FILE           Explicit file source (same as FILE).
                          
                          Supported file extensions:
                            .json               JSON
                            .yaml, .yml         YAML
                            .env                dotenv
                          
                          Sources are applied in the order provided; later values override earlier ones.
                          A default is always applied, regardless of whether --config is specified. (default env=IF_)
      --log-level level   Log level [debug info warn error dpanic panic fatal] (default info)
```

Closes #311


smira (approved):

smira:
I think this is very 🆒 

Probably should add a big warning to `hack/release.toml`

PR #342, opened 2025-12-17
feat: allow using image GET/HEAD API by the JS code on any domains

Unix4ever:
Changing some things to make Omni image factory UI interaction easier.

### DIFF
--- a/internal/frontend/http/http.go
+++ b/internal/frontend/http/http.go
@@ -19,6 +19,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/julienschmidt/httprouter"
 	"github.com/nicksnyder/go-i18n/v2/i18n"
+	"github.com/rs/cors"
 	"github.com/siderolabs/gen/ensure"
 	"github.com/siderolabs/gen/xerrors"
 	metrics "github.com/slok/go-http-metrics/metrics/prometheus"
@@ -62,6 +63,7 @@ type Options struct {
 	InstallerInternalRepository      name.Repository
 	InstallerExternalRepository      name.Repository
 	MetricsNamespace                 string
+	AllowedOrigins                   []string
 	RemoteOptions                    []remote.Option
 	RegistryRefreshInterval          time.Duration
 	ProxyInstallerInternalRepository bool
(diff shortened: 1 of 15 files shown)


internal/frontend/http/image.go
```diff
@@ -81,6 +81,7 @@ func (f *Frontend) handleImage(ctx context.Context, w http.ResponseWriter, r *ht
 	}
 
 	w.Header().Set("Content-Disposition", fmt.Sprintf(`attachment; filename="%s"`, path))
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
smira: I think this is only part of the story, you need to implement `OPTIONS` first and probably should inject this for all GET APIs. 

And of course it should be covered with tests

Unix4ever: We don't need it for all GET APIs though

internal/integration/download_test.go
```diff
@@ -140,6 +140,33 @@ func downloadAssetAndMatchSize(ctx context.Context, t *testing.T, baseURL string
...
+		return resp
+	}
+
+	for _, method := range []string{
+		http.MethodOptions, http.MethodGet, http.MethodHead,
+	} {
+		resp := doRequest(method)
+
+		assert.Equal(t, "*", resp.Header.Get("Access-Control-Allow-Domains"))
+		assert.Equal(t, "GET, HEAD, OPTIONS", resp.Header.Get("Access-Control-Allow-Methods"))
```
smira: does it work? I think it doesn't send this for non-OPTIONS call

internal/frontend/http/http.go
```diff
@@ -166,7 +167,16 @@ func NewFrontend(
...
 func (f *Frontend) Handler() http.Handler {
-	return f.router
+	return cors.New(cors.Options{
+		AllowedOrigins: []string{"*"},
+		AllowedMethods: []string{
+			http.MethodHead,
+			http.MethodGet,
+			http.MethodOptions,
+		},
+		AllowedHeaders: []string{"Authorization"},
```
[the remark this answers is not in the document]
Unix4ever: I think with * it's not possible to set AllowCredentials

PR #344, opened 2025-12-26
fix: drop sbc board support

frezbo:
Drop all SBC board support.

Fixes: https://github.com/siderolabs/talos/issues/12492


internal/frontend/http/ui.go
```diff
@@ -362,18 +362,26 @@ func (p WizardParams) URLValues() url.Values {
...
-	versions, err := f.getTalosVersions(ctx, params.SelectedVersion)
+	versions, err := f.getTalosVersions(ctx, params.SelectedVersion, params.Target)
 	if err != nil {
 		return "", nil, nil, err
 	}
 
+	var message string
+
+	if params.Target == TargetSBC {
+		message = "Only Talos 1.7+ is supported with SBC's"
```
smira: I think this should be pushed down to the template and translated (we can make same condition in the template)

smira (approved):

PR #345, opened 2025-12-29
fix: don't enforce bundle verified

smira:
This is not set by cosign when verifying using raw certificate verification.

The problem was introduced #337 (never tagged as a release).

### DIFF
--- a/internal/artifacts/internal/image/verify.go
+++ b/internal/artifacts/internal/image/verify.go
@@ -62,7 +62,7 @@ func verifyLegacySignature(ctx context.Context, digestRef name.Reference, ivo co
 	ivo.NewBundleFormat = false
 
 	_, bundleVerified, err := cosign.VerifyImageSignatures(ctx, digestRef, &ivo)
-	if err == nil && bundleVerified {
+	if err == nil {
 		// determine verification method
 		var verificationMethod string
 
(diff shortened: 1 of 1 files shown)


kevintijssen (approved):

frezbo:
> so this is for self-signed right?

but then i think we should distinguish between self-signed and signed by rekor, since we could potentially allow a non-signed image

smira:
> > so this is for self-signed right?
> 
> but then i think we should distinguish between self-signed and signed by rekor, since we could potentially allow a non-signed image

I believe it still returns an error if it can't verify a public signature (that's my understanding). It just says that with a self-signed cert it can't verify

smira:
> > > > so this is for self-signed right?
> > > 
> > > 
> > > but then i think we should distinguish between self-signed and signed by rekor, since we could potentially allow a non-signed image
> > 
> > I believe it still returns an error if it can't verify a public signature (that's my understanding). It just says that with a self-signed cert it can't verify
> 
> it should just concerned if we have a case where err is nil but not bundleverified 

Yes, but that's what happens with a custom public key verifier, it never sets bundle verified.

In fact, I reverted my changes in previous PR, this is how the code flow looks like in v0.9.0

frezbo (approved):

frezbo:
we can use this actually:

```
func validateNoSignatureFound(err error) bool {
	var noSignatureFoundErr *cosign.ErrNoSignaturesFound

	return errors.As(err, &noSignatureFoundErr)
}

func validateNoMatchingAttestations(err error) bool {
	var noAttestationFoundErr *cosign.ErrNoMatchingAttestations

	return errors.As(err, &noAttestationFoundErr)
}
```

frezbo:
> > we can use this actually:
> 
> Not sure how that would help... I mean we can improve error reporting, but that won't change how it works, right?

yeh, this is fine then

PR #348, opened 2026-01-08
fix: running image-factory rootless

frezbo:
Set `TUF_ROOT` so that it uses a writable location.


smira:
Any context on this change? How did it work before? Do the users running Image Factory need to set this env variable?

PR #349, opened 2026-01-09
chore: update `developing.md`, refactor kres

laurazard:
`RUN_TESTS` was being used for the `integration-direct`, `integration-s3`, `integration-cdn`, and other tests. This meant that, for all of these, the makefile has one default value (`TestIntegrationCDN`, previously), making running these targets confusing (`make integration-direct`) would run `TestIntegrationCDN` unless `RUN_TESTS=TestIntegrationDirect`
was set.

This patch splits `RUN_TESTS` into `RUN_TESTS_DIRECT`, `RUN_TESTS_CDN`, `RUN_TESTS_S3`, etc., which allows each of these to have a sane default in the Makefile.

Also update `docs/developing.md`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,13 +83,8 @@ TOOLCHAIN ?= docker.io/golang:1.25-alpine
 # extra variables
 
 PKGS_PREFIX ?= ghcr.io/siderolabs
-<<<<<<< HEAD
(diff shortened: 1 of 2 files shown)


frezbo (approved):
🆒

PR #359, opened 2026-01-21
feat: support copying to clipboard

### LINKED ISSUE #339: Add clipboard copy buttons on final page
Items that are intended to be copied should have a copy to clipboard button to avoid partial text selection

Image

smira:
Fixes #339


internal/frontend/http/templates/common.html
```diff
@@ -127,3 +127,16 @@ <h1 class="mb-4 mt-2 text-2xl font-bold">{{ . }}</h1>
 {{ end }}
 {{ end }}
 {{ end }}
+
+{{ define "copyable-text"}}
+<div class="inline-flex items-center gap-2 bg-slate-200 dark:bg-slate-700 p-1">
+        <span class="font-mono">
+                {{ . }}
+        </span>
+        <button type="button" class="p-0.5 text-gray-600 hover:text-gray-900 dark:text-gray-400 dark:hover:text-gray-100 rounded border border-gray-300 dark:border-gray-600 hover:bg-gray-100 dark:hover:bg-gray-700 [&[data-copied]]:bg-green-100 dark:[&[data-copied]]:bg-green-700 [&[data-copied]]:text-green-600 dark:[&[data-copied]]:text-green-300 [&[data-copied]]:border-green-300 dark:[&[data-copied]]:border-green-700"
```
Slessi: could also add a `transition-colors` class so it has a nice fade in/out when changing colours to confirm copying

smira: ❤️  thanks

PR #361, opened 2026-01-22
fix: clarify bootloader selection

### LINKED ISSUE #340: Bootloader selection explaination
We should make it clearer which bootloader to use

Image

Maybe instead of labeling them by what bootloader is being used they should be listed by which initialization system is being used. eg

* auto <-- how is this determined and is it different from dual-boot?
* UEFI: Modern hardware with recommended initialization. Uses systemd-boot (sd-boot).
* BIOS: Legacy hardware initialization system. Uses grub and cannot use secureboot.

smira:
Fixes #340


smira:
image

PR #410, opened 2026-03-06
feat: add SPDX frontend

### LINKED ISSUE #284: Publish spdx files with generated images
When a user generates a custom image with extensions we should publish an spdx file they can download and import into 3rd party tools for internal compliance and CVE verification.

Currently, if someone wants to verify CVEs from a 3rd party tool they can download the generic SBOM from github (without extensions) or they can read the spdx file from a running Talos system. Reading the spdx file from a running system wouldn't work for a system connected to Omni because of limits on the Talos API.

shanduur:
image

Fixes #284


cmd/image-factory/cmd/options.go
```diff
@@ -369,6 +372,57 @@ type ComponentsOptions struct {
...
+	if s.Raw == "time.Now()" || s.Raw == "now" {
+		return s.Raw
+	}
+
+	return s.Time().Format(time.RFC3339)
+}
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface for SPDXGeneratedTime.
+func (s *SPDXGeneratedTime) UnmarshalText(text []byte) error {
+	input := string(text)
```
frezbo: i think it's better if set to empty string just assume `time.Now()`

cmd/image-factory/cmd/options_test.go
```diff
@@ -6,12 +6,141 @@ package cmd_test
...
+			actual := cmd.SPDXGeneratedTime{
+				Time: func() time.Time { return now },
+			}
+
+			err := actual.UnmarshalText([]byte(input))
+			require.NoError(t, err)
+
+			t1 := actual.Time()
+
+			time.Sleep(time.Second)
```
frezbo: we should wrap with `testing/synctest` see https://go.dev/blog/testing-time

cmd/image-factory/cmd/options.go
```diff
@@ -369,6 +372,58 @@ type ComponentsOptions struct {
 	Talosctl string `koanf:"talosctl"`
 }
 
+// SPDXOptions configures the generation of SPDX documents for built images.
+type SPDXOptions struct {
+	// SPDXGeneratedTime configures the timestamp included in generated SPDX documents.
+	// If not set or set to empty string, it will use the current time at generation.
+	// Otherwise, it should be a valid RFC3339 timestamp string.
+	SPDXGeneratedTime SPDXGeneratedTime `koanf:"generatedTime"`
```
smira: I wonder if we should instead use the timestamp coming from Talos core SPDX file? 

I understand we might want to make it configurable, but using the current timestamp will make it not reproducible, and any fixed timestamp makes little sense for any SPDX.

I believe we already set the Talos SPDX timestamp to the time of the commit, so for a release it will be release timestamp.

enterprise/spdx/builder/spdx.go
```diff
@@ -0,0 +1,180 @@
...
+	return bytes.NewReader(buf.Bytes()), int64(buf.Len()), nil
+}
+
+// mergeDocument merges a source SPDX document into the target merged document.
+//
+// All element IDs from the source document are prefixed with the source
+// identifier to prevent collisions between documents. DESCRIBES relationships
+// from the source document are rewritten to reference the merged document's
+// SPDXRef-DOCUMENT.
+func mergeDocument(merged, source *spdx.Document, sourceID string) {
```
smira, line 104: we also have [this](https://github.com/siderolabs/talos/blob/main/hack/sbom.sh) in Talos which I think does the same thing (merges multiple docs) but it also uses some replacements from @dsseng.

I wonder what @dsseng thinks about this part of code

internal/frontend/http/templates/wizard-final.html
```diff
@@ -128,6 +128,13 @@ <h2>{{ t .Localizer "final.extra" }}</h2>
                 {{ end }}
         </dl>
         {{ end }}
+
+        {{ if and .Enterprise .SBOMAvailable }}
```
smira: we should probably make a bit of a marketing move here, and show the SPDX link in a disabled way if !.Enterprise, saying that it's an Enterprise-only feature

internal/frontend/http/http.go
```diff
@@ -116,6 +118,22 @@ func NewFrontend(
 		registrator(path, httproutermiddleware.Handler(path, frontend.wrapper(handler), mdlw))
 	}
 
+	// enterprise
+	for _, enterpriseRoute := range enterprisePlugins {
```
smira, line 122: I love the enterprise plugins idea

internal/frontend/http/ui.go
```diff
@@ -617,6 +620,10 @@ func (f *Frontend) wizardFinal(ctx context.Context, params WizardParams) (string
 			TroubleshootingGuideAvailable: talosVersion.GTE(semver.MustParse("1.6.0")),
 			ProductionGuideAvailable:      talosVersion.GTE(semver.MustParse("1.5.0")),
 			TalosctlAvailable:             talosVersion.GTE(semver.MustParse("1.11.0-alpha.3")),
+			SBOMAvailable:                 talosVersion.GTE(semver.MustParse("1.12.0-alpha.0")),
```
smira: should probably be 1.11.0 (this is when we got SBOMs)

smira (commented):
I like it a lot, I left some notes.

We also need to update README.md for the new SPDX url, but keep a note that it's Enterprise only

enterprise/spdx/builder/spdx.go
dsseng: Yes, we still need the replacement there if using, it needs deterministic SPDX patch that's stuck in review for the time being

In Talos I implemented it using Syft's ability to pick up SPDX files during scanning and automatically adding them as dependencies. It should also work here I guess, making the new top-level document for (Talos + extensions), which includes Talos with given dependencies, and extensions.

I haven't yet run and tried this code, https://apps.rancher.io/sbom-viewer is the go-to tool I use for inspecting SBOMs so you might try it

cmd/image-factory/cmd/options.go
shanduur: Removed this param.

cmd/image-factory/cmd/options.go
shanduur: Makes sense, fixed!

internal/frontend/http/templates/wizard-final.html
shanduur: Done!

smira (approved):
🆒

PR #420, opened 2026-04-16
feat: add checksum frontend

### LINKED ISSUE #283: Create shas for download assets
When someone generates a downloadable image (e.g. iso) they want a way to verify their download with a sha. We should generate and publish a sha for assets that need to be manually downloaded in the image factory UI.

shanduur:
Add enterprise-only checksum endpoint: appending .sha256 or .sha512
to any /image/ path returns a single-line checksum file instead of
the asset. Algorithm is selected from the suffix; the Checksummer
interface takes the suffix so no circular import is needed.

Wizard UI shows (sha256) and (sha512) links per download button;
non-enterprise builds show a localized (checksums) tooltip.
Integration tests cover both algorithms (GET, HEAD, validate,
reproducibility, error cases).

Screenshot 2026-04-16 at 11 17 20
Screenshot 2026-04-16 at 11 20 35

Closes #283


cmd/image-factory/cmd/service.go
```diff
@@ -51,8 +51,6 @@ import (
 )
 
 // RunFactory runs the image factory with specified options.
-//
-//nolint:gocyclo,cyclop
```
shanduur, line 55: Refactored to drop the `gocyclo` and `cyclop` as well few others linters that might have complained soon-ish.

enterprise/checksum/builder/builder.go
```diff
@@ -0,0 +1,185 @@
...
+	// Check cache first
+	if err := b.storage.Head(ctx, profileHash); err == nil {
+		b.logger.Debug("checksum cache hit", zap.String("profileHash", profileHash))
+
+		return b.storage.Get(ctx, profileHash)
+	}
+
+	// Build using singleflight to prevent duplicate work
+	resultCh := b.sf.DoChan(profileHash, func() (any, error) {
+		return b.buildChecksum(schematicID, versionTag, path, profileHash) //nolint:contextcheck
```
smira: I wonder if it'd be a bit more reliable if we do SHA512 each time, so we don't cache the actual SHA512, but we used cached/built asset? and can we potentially reduce code duplication a bit by making the build code still same for all paths, but just checksum is under `enterprise/` ? 

The reason I think about caching both the asset and checksum separately, is that there is a risk of them getting out of sync

internal/frontend/http/ui.go
```diff
@@ -622,8 +623,9 @@ func (f *Frontend) wizardFinal(ctx context.Context, params WizardParams) (string
 			TalosctlAvailable:             talosVersion.GTE(semver.MustParse("1.11.0-alpha.3")),
 			SBOMAvailable:                 talosVersion.GTE(semver.MustParse("1.11.0")),
 
-			Enterprise:  enterprise.Enabled(),
-			SPDXBaseURL: f.options.ExternalURL.JoinPath("spdx", schematicID, version, params.Arch),
```
smira, line 626: we need to document new path in README please

enterprise/checksum/builder/builder.go
smira: I think this would simplify the code if we could "call" into regular "asset build" code, and once it produces the result, we just hook into it to do SHA512 and return the response?

shanduur: Yeah, makes sense.

internal/frontend/http/ui.go
shanduur: Fixed.

docs/api.md
```diff
@@ -15,6 +15,20 @@ grype sbom:response.spdx.json
 
 SPDX bundles are available for Talos versions **v1.11.0** and later.
 
+### `GET /checksum/:schematic/:version/:path`
```
smira: I _wonder_ if this should be `GET /:schematic/:version/:path.md5sum` so that downloaded file will have a proper extension, otherwise md5sum file path would be same as downloaded asset

.kres.yaml
```diff
@@ -953,3 +953,12 @@ spec:
       runnerGroup: large
     environment:
       REGISTRY: registry.dev.siderolabs.io
+---
+kind: golang.UnitTests
+spec:
+  extraArgs: "-tags=enterprise"
```
smira, line 942: 👍

smira (approved):
my only comment is about the path to keep it consistent

docs/api.md
shanduur: This is handled by content disposition header - it sets the filename. CURL respects it when `-J` flag is set, and `wget` when `--content-disposition` is set. Browsers follow it by default.

frezbo: hmm we should document it, since `-J` is not a commonly used curl flag

shanduur: I've added required `.sha512` suffix.

smira: yep, that's why I think if we put it into the path, it would work 100% across the board

we could split it up early, and if we see it, set a flag which invokes the enterprise frontend

smira: and less code duplication in general - once we are ready to send the download body, we just invoke the checksummer instead, with either `io.Reader` or a redirect URL (which we need to download and checksum)

smira: what I was trying to say is to drop the `/checksum/` prefix as well, and have it handled via the current download frontend most of the time... if that makes sense? the suffix `.sha512sum` is enough to understand what is going on?

shanduur: I've done a rewrite, it's working as expected now.

PR #422, opened 2026-04-16
feat: upgrade tailwind to v4

frezbo:
Upgrade tailwind to v4, switch to bun.lock

Also fix check-dirty step in CI and cleanup helm values update.


bun.lock
Slessi, line 1: pls no 🙏 Bun is honestly not yet as mature as Node. For Omni I had to go back to node because of some weird [issues](https://github.com/siderolabs/omni/issues/1398).

If there isn't a need can we stick to Node? image-factory is currently so simple that Bun would be fine, but if we wanna migrate over to a vite/vue setup to share components and designs from Omni, i'd rather not run into weird surprises when the setup gets more complex :<

package.json
```diff
@@ -1,7 +1,10 @@
 {
+  "devDependencies": {
+    "@tailwindcss/cli": "^4.0.0"
+  },
   "dependencies": {
     "@tailwindcss/forms": "^0.5.6",
     "@tailwindcss/typography": "^0.5.10",
```
Slessi: update these plugins as well?

internal/frontend/http/css/input.css
Slessi, line 1: you should delete the old config file, `tailwind.config.js`

internal/frontend/http/css/input.css
```diff
@@ -1,3 +1,6 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss";
+
+@source "../templates/**/*.html";
```
Slessi, line 3: before it was `*.{html,js}` but maybe this is fine, dont see any js files using tailind

Slessi (commented):
I havent had luck getting factory to run locally to validate the styles are still ok (the list of changes for v4 is a bit too big to statically review only) so I'll rely on you for that :D 

But code looks ok ✅ just some small comments

frezbo:
> I havent had luck getting factory to run locally to validate the styles are still ok (the list of changes for v4 is a bit too big to statically review only) so I'll rely on you for that :D
> 
> But code looks ok ✅ just some small comments

`make docker-compose-up PLATFORM=linux/arm64 REGISTRY=127.0.0.1:5010` works fine, it brings up full working factory

bun.lock
frezbo: yeh I can use node, it was just copied over from old kres

smira: it's not even in the actual code path, it's only used to compile Tailwind during the build, no JS stuff in the runtime

hack/dev/compose.yaml
```diff
@@ -17,7 +17,7 @@ services:
       - REGISTRY_STORAGE_DELETE_ENABLED=true
     restart: unless-stopped
   registry.ghcr.io:
-    image: registry:2.8
```
smira, line 20: is this intended? does it list tags correctly?

smira (approved):

hack/dev/compose.yaml
frezbo: yeh, it got fixed, the PR got merged yesterday https://github.com/distribution/distribution/pull/4846

Slessi (approved):

PR #423, opened 2026-04-16
feat: implement authentication support

### LINKED ISSUE #285: Add authentication to image factory
We'll be creating a new instance of the image factory for paying customers with additional features (eg FIPS) that should be only available after a customer has logged in and if they have an entitlement for the product. This may also be used to distribute extensions that are not free or legal for us to distribute (eg portworx).

This authentication will need to be connected to the customer account in stripe so they can use the same login from stripe and can be managed as one customer on the back end.

The image factory should still have unauthenticated options (factory.talos.dev isn't going away) and people will still need to be able to self-host it.

We will need a way for customers to do basic user management (similar to Omni) and for a way to generate programmatic authentication tokens for machines (talos nodes) so they can automatically update and fetch images.

shanduur:

Any access to the schematic requires the user to be authenticated
before access.

Moreover, any schematic stores the owner in the schematic, so each
schematic becomes private (owned by the user which created it).

Authentication is configured using a set of usernames and keys
associates with each user (API key).

Screenshot 2026-04-20 at 10 59 22
Screenshot 2026-04-20 at 13 49 21
Screenshot 2026-04-20 at 13 49 31

Closes https://github.com/siderolabs/image-factory/issues/285
Closes https://github.com/siderolabs/image-factory/pull/346

Co-authored-by: Andrey Smirnov <andrey.smirnov@siderolabs.com>

### DIFF
--- a/cmd/image-factory/cmd/options.go
+++ b/cmd/image-factory/cmd/options.go
@@ -5,12 +5,15 @@
 package cmd
 
 import (
+	"errors"
 	"strings"
 	"time"
+
+	"github.com/hashicorp/go-multierror"
 )
 
 // Options configures the behavior of the image factory.
-type Options struct {
+type Options struct { //nolint:govet // keeping order for semantic clarity
 	// HTTP configuration for the image factory frontend.
 	HTTP HTTPOptions `koanf:"http"`
 
@@ -31,6 +34,11 @@ type Options struct {
 
 	// Artifacts defines names and references for various images used by the factory.
 	Artifacts ArtifactsOptions `koanf:"artifacts"`
+
+	// Authentication settings.
+	//
+	// Note: only available in the Enterprise edition.
+	Authentication AuthenticationOptions `koanf:"authentication"`
 }
 
 // AssetBuilderOptions contains settings for building assets.
@@ -246,6 +254,9 @@ type S3CacheOptions struct {
 
 	// Enabled enables S3 cache.
 	Enabled bool `koanf:"enabled"`
+
+	// PresignedURLTTL is the duration for which presigned URLs are valid.
+	PresignedURLTTL time.Duration `koanf:"presignedURLTTL"`
 }
 
 // SchematicCacheOptions configures caching of schematic blobs.
@@ -395,6 +406,34 @@ type ComponentsOptions struct {
 	Talosctl string `koanf:"talosctl"`
 }
 
+// AuthenticationOptions holds authentication settings.
+type AuthenticationOptions struct { //nolint:govet // keeping order for semantic clarity
+	// Enabled enables authentication.
+	Enabled bool `koanf:"enabled"`
+	// HTPasswdPath is the path to the htpasswd file containing user credentials.
+	//
+	// The file follows the standard htpasswd format (username:bcrypt_hash, one per line).
+	// Multiple entries with the same username are supported, allowing multiple API keys per user.
+	// Only bcrypt hashes ($2y$/$2a$/$2b$) are accepted.
+	//
+	// It is required if authentication is enabled.
+	HTPasswdPath string `koanf:"htpasswdPath"`
+}
+
+// Validate checks for invalid or conflicting configuration.
+func (o *Options) Validate() error {
+	var err error
+
+	if o.Authentication.Enabled && o.Cache.CDN.Enabled {
+		err = multierror.Append(err,
+			errors.New("CDN cache cannot be enabled when authentication is enabled: "+
+				"CDN would serve assets without authentication, effectively leaking secrets",
+			))
+	}
+
+	return err
+}
+
 // DefaultOptions are the default options.
 var DefaultOptions = Options{
 	HTTP: HTTPOptions{
@@ -422,7 +461,8 @@ var DefaultOptions = Options{
 			Repository: "cache",
 		},
 		S3: S3CacheOptions{
-			Bucket: "image-factory",
+			Bucket:          "image-factory",
+			PresignedURLTTL: time.Hour,
 		},
 		Schematic: SchematicCacheOptions{
 			Capacity:    100_000,
--- a/enterprise/auth/auth.go
+++ b/enterprise/auth/auth.go
@@ -1,0 +1,208 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+//go:build enterprise
+
+// Package auth provides authentication mechanisms.
+package auth
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"path/filepath"
+	"sync/atomic"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/julienschmidt/httprouter"
+	"github.com/siderolabs/gen/xerrors"
+	"go.uber.org/zap"
+	"golang.org/x/crypto/bcrypt"
+
+	schematicpkg "github.com/siderolabs/image-factory/pkg/schematic"
+)
+
+const reloadInterval = 30 * time.Second
+
+// NewProvider initializes a new authentication provider.
+// Call Run to start the background reload loop.
+func NewProvider(htpasswdPath string, logger *zap.Logger) (*Provider, error) {
+	// dummyHash is used as a timing equalizer in verifyCredentials. Without it,
+	// a non-existent username returns instantly while a valid username with a wrong
+	// password incurs full bcrypt latency, creating a timing oracle for username enumeration.
+	dummyHash, err := bcrypt.GenerateFromPassword([]byte("image-factory-auth-timing-dummy"), bcrypt.DefaultCost)
+	if err != nil {
+		return nil, fmt.Errorf("auth: failed to generate dummy timing hash: %w", err)
+	}
+
+	users, err := LoadHTPasswd(htpasswdPath)
+	if err != nil {
+		return nil, err
+	}
+
+	p := &Provider{
+		path:      htpasswdPath,
+		logger:    logger.With(zap.String("component", "auth-provider")),
+		dummyHash: dummyHash,
+	}
+
+	p.users.Store(&users)
+
+	return p, nil
+}
+
+// Provider is an authentication provider backed by an htpasswd file.
+type Provider struct {
+	users     atomic.Pointer[map[string][]string]
+	logger    *zap.Logger
+	path      string
+	dummyHash []byte
+}
+
+// Run starts the background file-watcher and reload loop.
+// It blocks until ctx is canceled.
+func (p *Provider) Run(ctx context.Context) error {
+	p.startWatcher(ctx)
+
+	return nil
+}
+
+func (p *Provider) startWatcher(ctx context.Context) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		p.logger.Warn("htpasswd: failed to create fsnotify watcher, using polling only", zap.Error(err))
+		p.pollLoop(ctx)
+
+		return
+	}
+
+	defer watcher.Close() //nolint:errcheck
+
+	// Watch the directory so we catch k8s ConfigMap symlink rotations.
+	dir := filepath.Dir(p.path)
+
+	if watchErr := watcher.Add(dir); watchErr != nil {
+		p.logger.Warn("htpasswd: failed to watch directory, using polling only",
+			zap.String("dir", dir), zap.Error(watchErr))
+		p.pollLoop(ctx)
+
+		return
+	}
+
+	ticker := time.NewTicker(reloadInterval)
+	defer ticker.Stop()
+
+	cleanPath := filepath.Clean(p.path)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+
+			if filepath.Clean(event.Name) == cleanPath &&
+				(event.Has(fsnotify.Write) || event.Has(fsnotify.Create)) {
+				p.reload()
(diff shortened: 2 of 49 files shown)


enterprise/auth/auth.go
```diff
@@ -0,0 +1,204 @@
...
+	schematicpkg "github.com/siderolabs/image-factory/pkg/schematic"
+)
+
+const reloadInterval = 30 * time.Second
+
+// dummyHash is computed once at startup and used as a timing equalizer in
+// verifyCredentials. Without it, a non-existent username returns instantly
+// while a valid username with a wrong password incurs full bcrypt latency,
+// creating a timing oracle for username enumeration.
+var dummyHash []byte
```
utkuozdemir: Nit: I'd put this on provider as a field to not introduce an init function.

enterprise/auth/auth.go
```diff
@@ -0,0 +1,204 @@
...
+		panic("auth: failed to generate dummy timing hash: " + err.Error())
+	}
+
+	dummyHash = hash
+}
+
+// NewProvider initializes a new authentication provider and starts a background
+// goroutine that reloads the htpasswd file when it changes.
+// The goroutine runs until ctx is canceled.
+func NewProvider(ctx context.Context, htpasswdPath string, logger *zap.Logger) (*provider, error) {
```
utkuozdemir: AFAIK returning unexported struct from an exported function is considered a bad practice.

enterprise/auth/auth.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	p := &provider{
+		path:   htpasswdPath,
+		logger: logger.With(zap.String("component", "auth-provider")),
+	}
+
+	p.users.Store(&users)
+
+	go p.startWatcher(ctx)
```
utkuozdemir: I would maybe suggest changing this to "leave concurrency to the caller", as suggested in Practical Go.

shanduur: Done!

enterprise/auth/auth.go
shanduur: Fixed!

enterprise/auth/auth.go
shanduur: Done, populated in constructor.

pkg/enterprise/enterprise.go
```diff
@@ -53,3 +54,19 @@ type ErrNotEnabledTag struct{}
...
 	WriteChecksum(ctx context.Context, w http.ResponseWriter, r *http.Request, reader io.ReadCloser, size int64, filename, suffix string) error
 }
+
+// AuthProvider defines an authentication provider.
+type AuthProvider interface {
+	// Run starts the background reload loop and blocks until ctx is canceled.
+	Run(ctx context.Context) error
+
+	Middleware(
+		func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error,
```
utkuozdemir: nit: might consider  giving a type alias to `func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error` for readability.

cmd/image-factory/cmd/options.go
```diff
@@ -395,6 +406,34 @@ type ComponentsOptions struct {
...
+
+	if o.Authentication.Enabled && o.Cache.CDN.Enabled {
+		err = multierror.Append(err,
+			errors.New("CDN cache cannot be enabled when authentication is enabled: "+
+				"CDN would serve assets without authentication, effectively leaking secrets",
+			))
+	}
+
+	return err
+}
```
utkuozdemir, line 435: I assume this style is intentional so if we need later to add another validation, we wouldn't miss to make it use errgroup, right?

enterprise/auth/auth_test.go
utkuozdemir, line 1: I can suggest adding tests for watch/reload logic as well.

utkuozdemir (approved):
LGTM overall, if I didn't miss anything 🚀

cmd/image-factory/cmd/options.go
shanduur: Yes. I used the `multierror.Append` as it presents the errors to user in a better way than `errors.Join` when stringified.

enterprise/auth/auth_test.go
shanduur: There is `testAuthReload`.

utkuozdemir: oh, sorry, missed that

cmd/image-factory/cmd/options.go
```diff
@@ -395,6 +406,34 @@ type ComponentsOptions struct {
...
+	//
+	// It is required if authentication is enabled.
+	HTPasswdPath string `koanf:"htpasswdPath"`
+}
+
+// Validate checks for invalid or conflicting configuration.
+func (o *Options) Validate() error {
+	var err error
+
+	if o.Authentication.Enabled && o.Cache.CDN.Enabled {
```
smira, line 427: I think we should not block this, this is more of CDN configuration, rather than a thing we should restrict on the Image Factory side.

What we should block (I guess) is that the installer registry should be always proxied via IF to ensure authentication.

enterprise/auth/auth.go
```diff
@@ -0,0 +1,208 @@
...
+		select {
+		case <-ctx.Done():
+			return
+
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+
+			if filepath.Clean(event.Name) == cleanPath &&
```
smira, line 111: I wonder if it works with `Secret` mapped as a volume, iirc there was a symlink created to the actual file which got updated by the kubelet on Secret change?

cmd/image-factory/cmd/options.go
shanduur: > I think we should not block this, this is more of CDN configuration, rather than a thing we should restrict on the Image Factory side.

I do think this should be secure by default - and unless we have a solution that proves there is a support for presigned URLs in CDN, I don't want to risk anyone misconfiguring this (either on our or customer side) and leaking sensitive data.

> What we should block (I guess) is that the installer registry should be always proxied via IF to ensure authentication.

Agree, I'll add fix for that.

enterprise/auth/auth.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Handler is the type of HTTP handlers used by the enterprise frontend.
+type Handler = func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error
+
+// Middleware implements enterprise.AuthProvider.
+func (provider *Provider) Middleware(handler Handler) Handler {
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error {
+		username, password, ok := r.BasicAuth()
+		if !ok {
+			return xerrors.NewTagged[schematicpkg.RequiresAuthenticationTag](errors.New("authentication required"))
```
smira, line 163: I'm a bit confused here - if we have no auth in the request, should we just continue in the middleware without injecting credentials?

enterprise/spdx/frontend.go
```diff
@@ -70,8 +103,13 @@ func (f *Frontend) Methods() []string {
 func (f *Frontend) Handle(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error {
 	schematicID := p.ByName("schematic")
 
-	// Validate schematic exists
-	if _, err := f.schematicFactory.Get(ctx, schematicID); err != nil {
+	// Validate schematic exists and check ownership.
+	schematic, err := f.schematicFactory.Get(ctx, schematicID)
```
smira, line 107: I _think_ for defensive programming reasons we should inject authz check into the `schematicFactory.Get` - this implementation e.g. short-circuits if the schematic is not found, while it should return an error if the user is not authorized

internal/frontend/http/configuration.go
```diff
@@ -28,12 +28,18 @@ func (f *Frontend) handleSchematicCreate(ctx context.Context, w http.ResponseWri
...
 
-	cfg, err := schematic.Unmarshal(data)
+	schematic, err := schematicpkg.Unmarshal(data)
 	if err != nil {
 		return err
 	}
 
-	id, err := f.schematicFactory.Put(ctx, cfg)
+	if f.options.AuthProvider != nil {
+		if username, ok := f.options.AuthProvider.UsernameFromContext(ctx); ok {
```
smira, line 37: I don't get the flow here, if the AuthProvider is not nil, the username _should_ be in the request, otherwise it doesn't make any sense? There should be no schematic created without an owner in case if auth is enabled?

internal/frontend/http/http.go
```diff
@@ -137,43 +145,45 @@ func NewFrontend(
...
 
-	// talosctl
-	registerRoute(frontend.router.HEAD, "/talosctl/:version/:path", frontend.handleTalosctl)
-	registerRoute(frontend.router.GET, "/talosctl/:version/:path", frontend.handleTalosctl)
+	// talosctl - public
+	registerPublicRoute(frontend.router.HEAD, "/talosctl/:version/:path", frontend.handleTalosctl)
+	registerPublicRoute(frontend.router.GET, "/talosctl/:version/:path", frontend.handleTalosctl)
 
-	// UI
+	// UI - require auth (consistent with all other schematic-creating endpoints)
```
smira, line 186: UI should only require auth once it touches the schematic, I don't think we should enforce auth before we get to the schematic-sensitive operations

internal/frontend/http/http.go
```diff
@@ -267,6 +303,29 @@ func MatchError(err error, callback func(message string, code int)) (zapcore.Lev
 	return level, status
 }
 
+// checkOwnership verifies the context user owns the schematic (for auth-protected routes).
+// Returns nil if schematic has no owner or the authenticated user matches the owner.
+func (f *Frontend) checkOwnership(ctx context.Context, s *schematicpkg.Schematic) error {
+	if s.Owner == "" {
```
smira, line 309: this seems to be wrong, if the auth is enabled, there should be no schematic without an owner, so I think the conditions here are misplaced

internal/frontend/http/image.go
```diff
@@ -64,6 +66,10 @@ func (f *Frontend) handleImage(ctx context.Context, w http.ResponseWriter, r *ht
 		return err
 	}
 
+	if err = f.checkOwnership(ctx, schematic); err != nil {
```
smira, line 69: see my other comments - let's push all checks down to the schematic service

internal/frontend/http/image.go
```diff
@@ -107,7 +113,11 @@ func (f *Frontend) handleImage(ctx context.Context, w http.ResponseWriter, r *ht
 		return f.checksummer.WriteChecksum(ctx, w, r, reader, asset.Size(), filename, checksumSuffix)
 	}
 
-	if asset, ok := asset.(cache.RedirectableAsset); ok && !disableRedirect && r.Method != http.MethodHead {
+	// When auth is active, never redirect to S3/CDN - bytes must flow through the
```
smira, line 116: I don't think we should enforce it here, but rather for the installer

internal/frontend/http/pxe.go
```diff
@@ -76,6 +81,14 @@ func (f *Frontend) handlePXE(ctx context.Context, w http.ResponseWriter, _ *http
 		return err
 	}
 
+	// Embed credentials in image asset URLs so iPXE can authenticate when fetching kernel/initramfs.
+	imageBaseURL := f.options.ExternalPXEURL
+	if username, password, ok := r.BasicAuth(); ok {
```
smira, line 86: I think we should have a pre-condition here on auth provider, we should not propagate any HTTP basic auth down if the auth is not enabled

enterprise/auth/auth.go
shanduur: No, why? If there is no credentials and the middleware is applied, we must reject the request.

enterprise/spdx/frontend.go
shanduur: SPDX frontend is authenticated as well - if we reach this point, it means that the auth was provided.

Or do you mean that even if we are not authorized here, we should always return `404`, to not show that the schematic exists?

internal/frontend/http/http.go
shanduur: Because our UI is HTMX-based SPA I don't think there is a way to trigger the authentication prompt directly from HTMX request.

cmd/image-factory/cmd/options.go
smira: > I do think this should be secure by default - and unless we have a solution that proves there is a support for presigned URLs in CDN, I don't want to risk anyone misconfiguring this (either on our or customer side) and leaking sensitive data.

We still should not block on that - we can warn, but we don't know if the CDN is secure or not.

Actually same issue with the installer registry - if it's backed by the CDN, same issue appears - it redirects to the CDN outside of IF control (even if proxied via IF). 

So we can warn/document that CDN in general should be protected from request replay by checking signatures, but we should not block.

enterprise/auth/auth.go
smira: yeah, I got it reading the code further, I assumed it's always in the path, ignore me :)

enterprise/spdx/frontend.go
smira: SPDX frontend depends on the schematic, so it would naturally depend on auth.

My point was more that if there is no auth in the context, `schematicFactory.Get` should return "not authenticated" early, no matter if the schematic exists or not (to prevent enumeration).

While if the user is authenticated, it can return 404 for the schematic which doesn't exist, or not authorized if the schematic's owner doesn't match, or the schematic on success.

internal/frontend/http/http.go
smira: yeah, discussed separately, let's keep it this way (as in the PR).

PR #424, opened 2026-04-16
feat: update Talos to v1.13.0-rc.0

smira:
Rename consistently to 'Image Factory Enterprise'.

Pass Talos name down to profile.

Pass the image factory name (and url) in the schematic's extension Author field.


internal/artifacts/schematic.go
```diff
@@ -62,13 +62,13 @@ func (m *Manager) GetSchematicExtension(ctx context.Context, versionTag string,
...
 // schematicExtension builds a "virtual" extension matching a specified schematic.
-func schematicExtension(schematicID string, schematicInfo []byte) (io.Reader, error) {
+func schematicExtension(schematicID string, schematicInfo []byte, baseURL string) (io.Reader, error) {
 	manifest := extensions.Manifest{
 		Version: "v1alpha1",
 		Metadata: extensions.Metadata{
 			Name:        constants.SchematicIDExtensionName,
 			Version:     schematicID,
-			Author:      "Image Factory",
+			Author:      constants.ImageFactoryName + " (" + baseURL + ")",
```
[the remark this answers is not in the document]
smira: it can't be empty

internal/artifacts/schematic.go
```diff
@@ -62,13 +62,13 @@ func (m *Manager) GetSchematicExtension(ctx context.Context, versionTag string,
 }
 
 // schematicExtension builds a "virtual" extension matching a specified schematic.
-func schematicExtension(schematicID string, schematicInfo []byte) (io.Reader, error) {
+func schematicExtension(schematicID string, schematicInfo []byte, baseURL string) (io.Reader, error) {
```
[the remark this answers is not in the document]
smira: it can't be empty

go.mod
```diff
@@ -2,6 +2,11 @@ module github.com/siderolabs/image-factory
 
 go 1.26.1
 
+// downgrade the go-containerregistry to workaround the security changes in
+// https://github.com/google/go-containerregistry/pull/2227/changes: it breaks
+// our extensions
```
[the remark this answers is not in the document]
smira: yes, we should remove it.

go.mod
```diff
@@ -2,6 +2,11 @@ module github.com/siderolabs/image-factory
...
+replace github.com/google/go-containerregistry => github.com/google/go-containerregistry v0.21.2
```
frezbo, line 8: I think we can drop it now

frezbo (approved):

go.mod
smira: I would be cautious about this, let's do a full test before we go this way.

smira: And anyways we need to support old extensions which still contain those symlinks and whatnot.

frezbo: > And anyways we need to support old extensions which still contain those symlinks and whatnot.

i think it was briefly only for the 1.13 alpha/beta versoion

smira: yep, so let's drop it after 1.13.0 if the testing in https://github.com/siderolabs/talos/pull/13140 goes well

smira: so let's remove it next release

PR #429, opened 2026-04-21
fix: centralize schematic ownership enforcement

shanduur:
Move ownership/auth checks from scattered frontend handlers into
  schematic.Factory.Get, which now accepts an OwnershipChecker. This
  eliminates duplicated checkOwnership methods across http and spdx
  frontends and ensures anonymous callers cannot probe schematic
  existence when auth is enabled.

  Also guard PXE credential embedding behind AuthProvider != nil so
  credentials are never propagated when auth is disabled.

### DIFF
--- a/internal/schematic/schematic.go
+++ b/internal/schematic/schematic.go
@@ -7,13 +7,20 @@ package schematic
 
 import (
 	"context"
+	"errors"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
 
 	"github.com/siderolabs/image-factory/internal/schematic/storage"
 	"github.com/siderolabs/image-factory/pkg/schematic"
 )
+
+// OwnershipChecker resolves the authenticated user from a context.
+type OwnershipChecker interface {
+	UsernameFromContext(ctx context.Context) (string, bool)
+}
 
 // Factory is the schematic factory.
 type Factory struct {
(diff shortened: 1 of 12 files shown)


internal/schematic/schematic.go
```diff
@@ -99,6 +106,42 @@ func (s *Factory) Get(ctx context.Context, id string) (*schematic.Schematic, err
...
+// Get retrieves the stored schematic and enforces ownership.
+//
+// If auth is non-nil and the caller is unauthenticated, RequiresAuthenticationTag is returned
+// even when the schematic is not found, to avoid leaking schematic existence to anonymous callers.
+func (s *Factory) Get(ctx context.Context, id string, auth OwnershipChecker) (*schematic.Schematic, error) {
+	sc, err := s.get(ctx, id)
+	if err != nil {
+		if auth != nil {
+			if _, ok := auth.UsernameFromContext(ctx); !ok {
+				return nil, xerrors.NewTagged[schematic.RequiresAuthenticationTag](errors.New("authentication required"))
```
smira: I think we don't return the actual error back (?) to the caller, we could wrap here `err` (so that it pops up in the log?)

internal/schematic/schematic.go
```diff
@@ -99,6 +106,42 @@ func (s *Factory) Get(ctx context.Context, id string) (*schematic.Schematic, err
...
+			}
+		}
+
+		return nil, err
+	}
+
+	if sc.Owner == "" {
```
smira: I think this would be more explicit

```suggestion
	if sc.Owner == "" && auth == nil {
```

smira (approved):

internal/schematic/schematic.go
shanduur: Done.

internal/schematic/schematic.go
shanduur: Right!

PR #431, opened 2026-04-24
feat: add /talosctl/:version endpoint to list downloadable talosctls

Slessi:
Add a `/talosctl/:version` endpoint which lists all downloadable `talosctl` binaries fro a given version.

Related to:
- https://github.com/siderolabs/omni/issues/2722

Requires:
- https://github.com/siderolabs/talos/pull/13201

<details>
<summary>
Sample response
</summary>

```json
[
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-darwin-amd64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-darwin-arm64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-freebsd-amd64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-freebsd-arm64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-linux-amd64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-linux-arm64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-linux-armv7",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-linux-riscv64",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-windows-amd64.exe",
  "http://localhost:8080/talosctl/1.13.0-rc.0/talosctl-windows-arm64.exe"
]
```
</details>


internal/frontend/http/talosctl.go
```diff
@@ -18,6 +19,42 @@ import (
...
+// handleTalosctls returns a map of Talos version to available talosctl binary filenames.
+func (f *Frontend) handleTalosctls(ctx context.Context, w http.ResponseWriter, _ *http.Request, _ httprouter.Params) error {
+	versions, err := f.artifactsManager.GetTalosVersions(ctx)
+	if err != nil {
+		return err
+	}
+
+	result := make(map[string][]string, len(versions))
+
+	for _, version := range versions {
```
smira: this is hell of a request to process, listing all versions of Talos going to the future.

smira (commented):
If you don't want to hardcode 1.11.0 as the min version, we can put to Talos's machinery imager/quirks, but I don't see any point in listing versions this way. 

talosctl is available for any version >= 1.11.0

internal/artifacts/manager.go
```diff
@@ -434,6 +434,15 @@ func (m *Manager) GetTalosctlTuples(ctx context.Context, versionString string) (
 		return tuples, nil
 	}
```
shanduur, line 436: Won't this cause the check bellow always return `nil, nil`?


The check in question:
```go
	if talosVersion.LT(semver.MustParse("1.11.0-alpha.3")) {
		return tuples, nil
	}
```

Slessi: yes, it will be a nil slice - that should be fine

internal/artifacts/manager.go
```diff
@@ -434,6 +434,15 @@ func (m *Manager) GetTalosctlTuples(ctx context.Context, versionString string) (
...
 
+	talosVersion, err := semver.ParseTolerant(versionString)
+	if err != nil {
+		return nil, err
+	}
+
+	if talosVersion.LT(semver.MustParse("1.11.0-alpha.3")) {
+		return tuples, nil
+	}
+
```
Slessi: @smira it is hardcoded to 1.11 here - i extracted this code from `ui.go` and moved it to the artifacts manager

internal/frontend/http/talosctl.go
Slessi: is this going to be the exact same list, for every single talos version? do you prefer if i drop this PR and just generate a list in omni with a fixed list of os/arch combinations?

i thought best to make factory the source of truth in case of supporting other possible combinations, or special enterprise builds, or anything else that can cause a list to change between versions

smira: I do not like the API which has basically unbounded complexity going forward, I don't see any point in having this list.

1.11.0 is a fixed point in time.  Available Talos versions is already available. Pull available version, filter `>= 1.11.0`, and you're done.

If you need to show downloads for a specific version, we have an API for that already - call it for a specific version and show the list.

internal/artifacts/manager.go
smira: Let's move it here: https://github.com/siderolabs/talos/blob/main/pkg/machinery/imager/quirks/quirks.go#L11

and share across Omni/Image Factory.

Slessi: Will update PR when merged

- https://github.com/siderolabs/talos/pull/13201

internal/frontend/http/talosctl.go
Slessi: The API is what i was using here, isn't exposed elsewhere - refactored PR to only list versions for one specific version at a time

go.mod
```diff
@@ -38,8 +37,8 @@ require (
 	github.com/siderolabs/go-blockdevice/v2 v2.0.28
 	github.com/siderolabs/go-debug v0.6.2
 	github.com/siderolabs/go-pointer v1.0.1
-	github.com/siderolabs/talos v1.13.0-rc.0
-	github.com/siderolabs/talos/pkg/machinery v1.13.0-rc.0
+	github.com/siderolabs/talos v1.13.0-alpha.2
+	github.com/siderolabs/talos/pkg/machinery v1.13.0-alpha.2.0.20260424113037-b43c3a124f6c
```
Slessi: temporary till PR is backported

internal/frontend/http/ui.go
```diff
@@ -608,12 +608,9 @@ func (f *Frontend) wizardFinal(ctx context.Context, params WizardParams) (string
 		secureBootInstallerImage = fmt.Sprintf("%s/%s-installer-secureboot/%s:%s", f.options.ExternalURL.Host, params.Platform, schematicID, version)
 	}
 
-	var talosctlTuples []artifacts.TalosctlTuple
-	if talosVersion.GTE(semver.MustParse("1.11.0-alpha.3")) {
```
Slessi, line 612: checked was moved into `f.artifactsManager.GetTalosctlTuples`

internal/frontend/http/http.go
```diff
@@ -180,6 +180,7 @@ func NewFrontend(
 	registerPublicRoute(frontend.router.GET, "/secureboot/signing-cert.pem", frontend.handleSecureBootSigningCert)
 
 	// talosctl - public
+	registerPublicRoute(frontend.router.GET, "/talosctl/:version", frontend.handleTalosctlList)
```
smira, line 183: three small things, please:

* update README.md with new API
* add new API to the client in this repo
* add an integration test (using the client) to assert that we get e.g. 1.11.0 & 1.12.0 in the list

go.mod
smira: v1.13.0 is around the corner now

smira (approved):

PR #432, opened 2026-04-24
feat(registry): resolve latest tag to stable version

shanduur:
"latest" now maps to latest non-prerelease version instead of external pass-through.

Part of https://github.com/siderolabs/talos/issues/13138

### DIFF
--- a/internal/frontend/http/registry.go
+++ b/internal/frontend/http/registry.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"slices"
 	"strings"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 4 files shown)


internal/frontend/http/registry.go
```diff
@@ -165,6 +166,14 @@ func (f *Frontend) handleManifest(ctx context.Context, w http.ResponseWriter, re
 		return err
 	}
 
+	// if the tag is "latest", replace it with latest known stable version to the factory.
+	if versionTag == "latest" {
+		versionTag, err = f.resolveLatest(ctx, schematicID)
```
smira, line 171: let's add to README.md & add an integration test for it!

smira (approved):

PR #433, opened 2026-04-26
Add french language

qjoly:
I noticed that the Factory website only supports three languages, so I thought it would be nice to add French (since the tech community really likes Talos) 😄


smira (approved):
thank you!

PR #434, opened 2026-04-27
feat: show schematic-id url parameter on the final wizard step

Orzelius:
This is useful as cli arguments and embedded config (to be added) can result in an extremely long url.
This will result in a 404 if the schematic is not registered.

### DIFF
--- a/internal/frontend/http/templates/wizard-cmdline.html
+++ b/internal/frontend/http/templates/wizard-cmdline.html
@@ -90,7 +90,7 @@
 
     <div class="flex gap-4">
         {{ template "back-button" (dict "Params"
-        "target,version,platform,board,arch,secureboot,bootloader,cmdline,cmdline-set,overlay-options,selected-extensions"
+        "target,version,platform,board,arch,secureboot,selected-extensions"
         "Localizer" .Localizer)
         }}
         {{ template "next-button" . }}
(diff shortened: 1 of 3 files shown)


smira (commented):
I think it's going in the right direction 👍

smira (commented):
there is a small regression now (I haven't checked why though):

if I move through pages like:

```
1 -> 2 -> 3 -> 4 -> (back) -> 3 -> 2 -> (forward) -> 3
```

IF correctly recovers settings when moving back, but if I go 2->3 (the last step), it should have forgot the selection on "3", but it still remembers it.

Orzelius:
> there is a small regression now (I haven't checked why though):
> 
> if I move through pages like:
> 
> ```
> 1 -> 2 -> 3 -> 4 -> (back) -> 3 -> 2 -> (forward) -> 3
> ```
> 
> IF correctly recovers settings when moving back, but if I go 2->3 (the last step), it should have forgot the selection on "3", but it still remembers it.

Is the desired behavior that stuff should not be remembered if moved back? Currently on main going `schematic ready -> Customization -> System Extensions -> Customization` still remembers the `Customization` setting somehow.

smira:
> Is the desired behavior that stuff should not be remembered if moved back? Currently on main going schematic ready -> Customization -> System Extensions -> Customization still remembers the Customization setting somehow

That was my initial goal, if you move forward, you should see "default" clean state always, it should only remember when going backwards.

Orzelius:
> That was my initial goal, if you move forward, you should see "default" clean state always, it should only remember when going backwards.

fixed the issue!

smira (approved):

PR #435, opened 2026-04-27
feat(enterprise): implement VEX endpoint

### LINKED ISSUE #427: support for VEX data
(Enterprise only).

Pull VEX data by Talos version: `https://factory.talos.dev/vex/v1.12.6/vex[-amd64].json` (no arch?)

Split `generate-vex` repository into source code (under BUSL), keep the source VEX yaml in `generate-vex`, publish the file inside `generate-vex` container image.

IF pulls `generate-vex` image, extracts source YAML and caches for some time (it pulls from `:latest`, and when it changes, it pulls new version).

IF uses VEX YAML + library -> generates vex.json, caches it for some short time, or uses VEX yaml in the cache ID.

shanduur:

Serves GET/HEAD /vex/:version/vex.json for Talos ≥ 1.13.0.
Pulls exploitability data from an OCI registry, generates a VEX
document via go-vex, and caches it in-memory with configurable TTL.

image

Closes #427

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 # THIS FILE WAS AUTOMATICALLY GENERATED, PLEASE DO NOT EDIT.
 #
-# Generated on 2026-04-22T20:13:43Z by kres 8299790.
+# Generated on 2026-05-05T11:14:47Z by kres 1762ab2.
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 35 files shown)


enterprise/vex/fetcher/fetcher.go
```diff
@@ -0,0 +1,116 @@
...
+	if f.insecure {
+		nameOpts = append(nameOpts, name.Insecure)
+	}
+
+	ref, err := name.NewTag(fmt.Sprintf("%s:%s", f.registry, version), nameOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing reference: %w", err)
+	}
+
+	imgDesc, err := f.puller.Get(ctx, ref)
```
smira: only one question that I haven't thought about (yet) - should we enforce the signature or not? 

I think for completeness sake, we should actually enforce the signature? 

We could also move this fetch part into common artifacts to share the signature verification? (and return e.g. `io.Reader` there and parse it here?

also related to this - should do we do common singleflight protection and cache the file at least for some time?

smira (approved):
🆒

PR #442, opened 2026-05-06
feat(enterprise): add vulnerability scanning endpoint

### LINKED ISSUE #428: support for security scan results
(Enterprise only).

Combine:

* SBOM
* VEX 
* vulnerability database

Delivers scan result in various formats (text, ...).

shanduur:

Fixes https://github.com/siderolabs/image-factory/issues/428

Implement Grype-backed vulnerability scanning for Talos SBOM with VEX
suppressions. Endpoint /scans/:schematic/:version/:arch/:report serves
scan results in multiple formats (json, table, sarif, cdx) with per-version
caching and configurable database URL for air-gapped deployments.

<details><summary>OSS</summary>
image
</details>

<details><summary>Enterprise (no-auth)</summary>
image
</details>

<details><summary>Enterprise (auth)</summary>
image
</details>

<details><summary>Scan details</summary>
image
</details>


cmd/image-factory/cmd/options.go
```diff
@@ -514,5 +527,9 @@ var DefaultOptions = Options{
 			},
 			CacheTTL: 15 * time.Minute,
 		},
+		Scanner: ScannerOptions{
+			DatabaseURL: "https://grype.anchore.io/databases",
+			CacheTTL:    15 * time.Minute,
```
frezbo: grype database is only updated every 24 hours, so having a 15 min TTL is just overkill

hack/dev/compose.yaml
```diff
@@ -3,11 +3,12 @@ configs:
     file: config.yaml
   if_signing_key:
     file: ../../_out/cache-signing-key.key
-  # if_htpasswd:
-  #   file: htpasswd
+  if_htpasswd:
```
frezbo, line 6: we can put static content here and avoid having a local file

frezbo (approved):
🆒

cmd/image-factory/cmd/options.go
shanduur: We are caching scan results. VEX data is also part of scan result, and it can change between DB updates.

enterprise/scanner/builder/builder.go
```diff
@@ -0,0 +1,355 @@
...
+	defer close(b.initDone)
+
+	scannerOpts := govexscanner.Options{ID: scannerID}
+
+	if opts.DatabaseURL != "" {
+		distConfig := distribution.DefaultConfig()
+		distConfig.LatestURL = opts.DatabaseURL
+
+		instConfig := installation.DefaultConfig(clio.Identification{Name: scannerID})
+		instConfig.DBRootDir = "/var/lib/grype"
```
smira, line 122: should this be somewhere in temp to allow easier emptydir mounts and skipping overlayfs overhead?

enterprise/scanner/builder/builder.go
```diff
@@ -0,0 +1,355 @@
...
+
+func cacheKey(schematicID, versionTag, arch string) string {
+	return schematicID + "|" + versionTag + "|" + arch
+}
+
+func (b *Builder) getCached(key string) (cachedScan, bool) {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+
+	item, ok := b.cache[key]
```
smira: the VEX number of cache keys is relatively small (just Talos versions), here number of keys is potentially big (with schematics included). 

I wonder if we should do LRU here, like we do for schematics in https://github.com/siderolabs/image-factory/blob/0d7cd1c8ad4c88f3e15d71746ae41838ff6113c3/internal/schematic/storage/cache/cache.go#L31-L32

smira (approved):

cmd/image-factory/cmd/options.go
frezbo: ah I thought this is the grype db refresh interval

enterprise/scanner/builder/builder.go
shanduur: This directory should be mounted as a volume - either EmptyDir or PVC (depending on the preference of users) and bypass overlay.

PVC will speed up the startup, while EmptyDir is simple option not requiring actual storage provider.

frezbo: i think it's fine to use emptyDir, grype db is small and we don;t restart factory often, it downloads pretty fast

enterprise/scanner/builder/builder.go
shanduur: Fixed!

PR #447, opened 2026-05-11
fix(config): validate early and sort SPDX deterministically

shanduur:
Config validation (externalURL, externalPXEURL) on startup -> catch
errors before runtime. Extracted buildDocumentNamespace for reuse.

Sort SPDX bundle files by source + filename -> fixes digest flips
when rebuilding with identical inputs (map iteration order variance).

Changes:
- Add Options.Validate() checks URLs valid/required
- Extract DocumentNamespace builder
- Sort files deterministic
- Add SPDX builder tests


enterprise/scanner/builder/builder.go
```diff
@@ -351,3 +361,37 @@ func renderReport(doc models.Document, s *sbom.SBOM, format govexscanner.ReportF
...
 	return data, nil
 }
+
+// NormalizeSBOMForScan rewrites the merged SPDX document so that any package
+// named with the build-flavor-specific TalosName ("Talos Enterprise" in the
+// Enterprise build) is renamed to the canonical vendor-neutral value.
+//
+// Grype derives the VEX product purl from the SBOM source name as
+// `pkg:generic/<lower(name)>@<ver>`. Upstream VEX statements target
+// `pkg:generic/talos@*`, so without normalization Enterprise scans miss every
```
frezbo: @smira should we publish separate vex for Talos enterprise, as in it could differ from standard talos in future 🤔  ?

I don't feel good about normalizing, feels like a hack to skip scanners

shanduur: Yeah, we probably should. However I'm not sure if the `Talos Enterprise` should actually be normalized in the Vex and Scanner to `Talos-Enterprise` - right now internally it is normalized by Grype to `talos enterprise` causing purl to change from `pkg:generic/talos@v1.13.0` in OSS to `pkg:generic/talos enterprise@v1.13.0` (which is not a valid purl I think? https://github.com/package-url/purl-spec). I'm not sure if this will actually have some consequences, but I'd prefer it to be not space separated - to make sure it's not breaking any tools others might have.

smira: I think it's good to have a different PURL for Talos Enterprise.

Right now VEX is identical for both Enterprise & OSS - might change in the future, but for now let's keep them in sync until we have a reason to split.

smira (approved):

PR #448, opened 2026-05-11
feat: add config embedding

Orzelius:
allow embedding initial machine config for enterprise image factory.

the download integration test needed some extra work to get it running with talos 1.13


internal/frontend/http/ui.go
```diff
@@ -663,6 +677,14 @@ func (f *Frontend) wizardFinal(ctx context.Context, params WizardParams) (string
 		return "", nil, nil, err
 	}
 
+	if params.EmbeddedConfig != "" {
+		if f.configEmbedder != nil {
+			f.configEmbedder.EmbedConfig(params.EmbeddedConfig, &requestedSchematic)
+		} else {
+			return "", nil, nil, fmt.Errorf("embedding machine configuration is only supported on Enterprise Image Factory. Is this the correct instance?")
```
Orzelius: The error will be logged, but the front end user will just see a "500 internal server error" which is not too helpful, but I think this is such a rare scenario that there's no point in trying to improve it for now

shanduur: We should wrap this with PaymentRequired error.

shanduur:
```go
xerrors.NewTaggedf[enterrors.NotEnabledTag]("enterprise not enabled: embedding machine configuration is not available")
```

internal/frontend/http/ui.go
```diff
@@ -493,6 +501,10 @@ func (p WizardParams) URLValues() url.Values {
 		values.Set("cmdline-set", "true")
 	}
 
+	if p.EmbeddedConfig != "" {
+		values.Set("embedded-config", p.EmbeddedConfig)
+	}
```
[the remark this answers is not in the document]
shanduur: Good bot.

We need a regression test for this as well.

pkg/schematic/schematic.go
```diff
@@ -32,6 +32,8 @@ type Schematic struct {
 
 // Customization represents the Talos image customization.
 type Customization struct {
+	// EmbeddedMachineConfiguration is the initial embedded machine configuration.
+	EmbeddedMachineConfiguration string `yaml:"embeddedMachineConfiguration,omitempty"`
```
smira, line 36: this should be documented as Enterprise only

smira (commented):
we also need:

* docs/api.md update for new schematic field (Enterprise only note)
* integration tests, preferably based on the ones which extract initramfs to verify that embedded config gets actually embedded (I believe we already verify extensions, and this should be an extension?)

smira (commented):
please stop pushing PRs to origin (push to your fork)

internal/frontend/http/locales/active.en.yaml
```diff
@@ -184,6 +184,18 @@
 - id: customization.overlay.skip
   translation: "If unsure, you can skip this step."
 
+- id: customization.embedded_config
+  translation: "Embedded machine configuration:"
+
+- id: customization.embedded_config.description
+  translation: "This configuration will be embedded into the image. For more details see the  "
```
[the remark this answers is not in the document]
Orzelius: The space is needed

Orzelius:
e2e test failure is unrelated

internal/profile/profile.go
```diff
@@ -332,6 +332,8 @@ func EnhanceFromSchematic(
 ) (profile.Profile, error) {
 	metricsOnce.Do(initMetrics)
 
+	prof.Customization.EmbeddedMachineConfiguration = schematic.Customization.EmbeddedMachineConfiguration
```
smira: in order to improve caching, we should move it under the if condition (the reason is that kernel/cmdline can't carry embedded configs, so no need to put it here)

```
if prof.Output.Kind != profile.OutKindCmdline && prof.Output.Kind != profile.OutKindKernel
```

and gate it under a quirk (and fail if the embedded config is not supported).

smira (commented):
probably also deserves a note in `hack/release.toml`

smira (approved):

docs/api.md
```diff
@@ -65,6 +65,11 @@ customization:
        # optional, include well-known UEFI certificates into auto-enrollment database (SecureBoot ISO only)
       includeWellKnownCertificates: true
     bootloader: sd-boot # optional, defaults to auto (bootloader chosen by imager), other options: dual-boot, grub
+    embeddedMachineConfiguration: | # optional, embedded machine configuration (YAML-encoded)
```
frezbo, line 68: can we put an example of using two documents, so it's not confusing

frezbo (approved):

docs/api.md
Orzelius: yeah, good idea

PR #449, opened 2026-05-13
docs: POST /schematics, clarify request body format

majabojarska:
Clarifies that the IF API expects a JSON-encoded request body, while the example is provided in YAML, to allow for comments explaining the individual fields (JSON doesn't support comments).

Also includes a yq YAML->JSON conversion command for convenience.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,7 +48,7 @@ Supported report formats:
 
 Create a new image schematic.
 
-The request body is a YAML (JSON) encoded schematic description:
+The request body is a **JSON-encoded** schematic description. To better explain the individual fields, here's a YAML example:
 
 ```yaml
 customization:
@@ -70,6 +70,11 @@ overlay: # optional
   name: rpi_generic # overlay name
   options: # optional, any valid yaml, depends on the overlay implementation
     data: "mydata"
+```
+
+You can easily convert from YAML to JSON with [yq](https://github.com/mikefarah/yq):
+```sh
+yq --input-format yaml --output-format json schematic.yaml 
 ```
 
 Output is a JSON object containing the schematic ID and the canonical schematic body as YAML:


majabojarska:
Closing, as upon a closer look, this was rooted in a misunderstanding. We were looking into why the following failed:
```sh
curl -X POST https://factory.talos.dev/schematics -d @/tmp/schematic.yaml
yaml: line 1, column 18: mapping values are not allowed in this context
talos on  feat/ci-validate-extensions-in-sync-with-talos +5 -5 [!?] via 🐳 orbstack
❯ cat /tmp/schematic.yaml
overlay:
    image: siderolabs/sbc-raspberrypi
    name: rpi_generic
```

We thought JSON was accepted instead, since the IF API accepted an equivalent JSON document just fine. This was a bad conclusion, as the API is built with first class YAML support ([1](https://github.com/siderolabs/image-factory/blob/3bccbe1286dde26f81fa084206a44e633f3c0324/internal/frontend/http/configuration.go#L22-L35), [2](https://github.com/siderolabs/image-factory/blob/3bccbe1286dde26f81fa084206a44e633f3c0324/pkg/schematic/schematic.go#L107-L110)). The reason JSON was accepted with the body provided through `-d`/`--data` is simple - JSON is not _as sensitive_ to whitespace as YAML. 

The actual culprit was using `-d` (short for `--data`), which per [the man page](https://curl.se/docs/manpage.html#:~:text=%2Dd%2C%20%2D%2Ddata%20%3Cdata):
> Posting data from a file named 'foobar' would thus be done with [--data](https://curl.se/docs/manpage.html#--data) @foobar. When [--data](https://curl.se/docs/manpage.html#--data) is told to read from a file like that, carriage returns, newlines and null bytes are stripped out

Sending the same request body, to the same endpoint, with `--data-binary` should preserve newlines, [per documentation](https://curl.se/docs/manpage.html#:~:text=%2D%2Ddata%2Dbinary%20%3Cdata):
> Data is posted in a similar manner as [--data](https://curl.se/docs/manpage.html#--data) does, except that newlines and carriage returns are preserved and conversions are never done.

And it does:
```sh
curl -X POST https://factory.talos.dev/schematics --data-binary @/home/majabojarska/if-json-yaml/schematic.yaml             18:11:05
{"id":"ee21ef4a5ef808a9b7484cc0dda0f25075021691c8c09a276591eedb638ea1f9","schematic":"overlay:\n    image: siderolabs/sbc-raspberrypi\n    name: rpi_generic\ncustomization: {}\n"}
```

---

Not a bug.

PR #457, opened 2026-05-20
feat: move SPDX cache to enterprise options

shanduur:
Migrate SPDX document cache configuration from global `Cache.OCI`
to enterprise-specific `Enterprise.SPDX.Cache` for better separation
of concerns and to enable per-component cache configurations.


smira (approved):

frezbo (approved):
might do a rekres again

PR #467, opened 2026-05-27
feat: allow excluding Talos releases

shanduur:
In case of broken Talos release, we keep serving it in the UI/API. This
can lead to issues when customers download broken release and try to run
it. We should be able to exclude the broken versions without creating
new release of Image Factory, purely by specifying them in config.

For API completeness, we will also serve /versions?broken=true endpoint
that lists the broken releases that should not be used.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -99,6 +99,16 @@ MinTalosVersion specifies the minimum supported Talos version for assets.
 
 ---
 
+### `build.brokenTalosVersions`
+
+- **Type:** `[]string`
+- **Env:** `BUILD_BROKENTALOSVERSIONS`
+
+BrokenTalosVersions lists Talos versions that should be considered broken and avoided when building assets.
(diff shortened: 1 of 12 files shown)


docs/configuration.md
```diff
@@ -99,6 +99,16 @@ MinTalosVersion specifies the minimum supported Talos version for assets.
 
 ---
 
+### `build.brokenTalosVersions`
+
+- **Type:** `[]string`
+- **Env:** `BUILD_BROKENTALOSVERSIONS`
```
smira, line 105: below it says `IF_BUILD_BROKEN_TALOSVERSIONS` ? which one is correct?

smira (approved):

frezbo (approved):

docs/configuration.md
shanduur: The `IF_` prefix is settable, can be disabled. We are only showing here the core path.

PR #472, opened 2026-06-01
feat: bump go-conainerregistry

frezbo:
feat: bump go-conainerregistry

Add options to configure debugging and setting default parallel jobs.

See: https://github.com/siderolabs/talos/pull/13462

Signed-off-by: Noel Georgi <git@frezbo.dev>


internal/remotewrap/remotewrap.go
```diff
@@ -20,6 +20,16 @@ import (
...
 
+// defaultRemoteJobs sets the parallelism for remote pull/push operations.
+//
+// go-containerregistry v0.21.x introduced a pull limiter that gates concurrent
+// blob fetches on this value (default 4), with tokens only released when the
+// layer reader is closed. The default of 4 throttles Image Factory, which
+// fans out many concurrent image/layer pulls through a small set of shared
+// pullers. Restore the effectively-unbounded concurrency we relied on before
+// the limiter existed; actual concurrency stays bounded by the connection pool.
+const defaultRemoteJobs = 64
```
shanduur: Do we need this configurable? Or is the default value right here?

frezbo: i just put a number, go-containerregistry sets as 4 by default, should I move this to configuration?

smira: I think it should be configurable... Probably a default value should be relatively low, but users should be allowed to control it

frezbo: what should be the default, the 4 from upstream makes CI slow, so that's not ideal at all

smira: CI is an edge case really, it does so many builds in parallel, not a real deployment.

I wonder if it also counts our _internal_ registries as well, that might make more sense

smira (approved):

PR #476, opened 2026-06-04
feat: add a hasher for spdx

frezbo:
Add a hasher similar to profile.Hash for spdx, this allows us to invalidate caches when we have an errata.

The current change to use hasher now has the effect that all existing ones are invalidated, since we dropped the `spdx-` prefix to align with how hashes are used for profiles too. So no errata are being added now.

### DIFF
--- a/enterprise/spdx/builder/hash.go
+++ b/enterprise/spdx/builder/hash.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+//go:build enterprise
+
+package builder
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+)
+
+// Hash returns a content hash describing the inputs and extraction logic that
+// produce an SPDX bundle. It is used directly as the OCI cache tag, so that
+// fixes to the SPDX extraction/merge logic can invalidate previously cached
+// bundles even though the schematic, version and architecture are unchanged.
+//
+// This mirrors internal/profile.Hash (whose output is likewise used as the
+// asset cache tag): the inputs are checksummed and then errata strings are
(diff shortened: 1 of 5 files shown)


enterprise/spdx/builder/hash.go
```diff
@@ -0,0 +1,43 @@
...
+// bundles even though the schematic, version and architecture are unchanged.
+//
+// This mirrors internal/profile.Hash (whose output is likewise used as the
+// asset cache tag): the inputs are checksummed and then errata strings are
+// mixed in, bumping the hash (and therefore the cache key) whenever a bug in a
+// previously cached SPDX bundle needs to be invalidated.
+//
+// Operators are expected to use distinct cache repositories for OSS vs
+// Enterprise deployments since the bundle content differs by build flavor.
+func Hash(schematicID, version, arch string) string {
```
smira, line 27: I would like us to move way forward on this one, and do proper caching for SBOMs.

If we just depend on schematic IDs, two schematics with e.g. just a different kernel args will not produce same SBOM ID.

Let's reconsider this, probably @shanduur can help here.

It _feels_ like we only need list of extensions from the schematic (plus overlay?), but let's do a proper tested hashing here to avoid recomputing the SPDX.

Same applies to scan results - if a scan is requested, if the SPDX hash is same, don't re-run the scan unless vulndb/VEX changed.

frezbo: > If we just depend on schematic IDs, two schematics with e.g. just a different kernel args will not produce same SBOM ID.

Ahhh that's bad already, we should rely on extensions only, overlays don't ship sboms, unless we want them to

shanduur: Yeah, schematicID is too fragile, we should cache on Version+Arch+Extensions.

shanduur:
We can close this, right?

frezbo:
> We can close this, right?

yes

PR #482, opened 2026-06-12
feat: add secureboot enrollKeys schematic option

mcanevet:
## What?

Add a `secureboot.enrollKeys` field to the schematic, mapping to the imager's `SDBootEnrollKeys` profile option (`off` / `manual` / `if-safe` / `force`) for the ISO and disk-image outputs. It controls systemd-boot's `secure-boot-enroll` setting in `loader.conf`. The field is `omitempty`, so schematics that don't set it produce an unchanged ID.

```yaml
customization:
  secureboot:
    enrollKeys: force
```

## Why?

The default `if-safe` auto-enrolls SecureBoot keys only inside a VM, so on bare-metal keys are never enrolled unattended. Network provisioning flows with no console operator (PXE, Tinkerbell, Metal³, …) need `force`, which enrolls once while the UEFI firmware is in setup mode — sd-boot's setup-mode gate prevents it from ever overwriting an already-enrolled platform key.

Follow-up to the imager flag (siderolabs/talos#13571), per siderolabs/talos#12568 where @smira said this "can [be] put to the Image Factory as well" (as an opt-in; default stays `if-safe`).

## Notes

- Mapped per output kind (ISO → `ISOOptions`, disk image → `ImageOptions`); `secure-boot-enroll` is meaningless for installer/UKI outputs, so it's ignored there.
- Tests added for the mapping (force → image/ISO, invalid value rejected, ignored without SecureBoot); existing golden profile and schematic-ID tests are unchanged.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -64,6 +64,9 @@ customization:
     secureboot: # optional, only applies to SecureBoot images
        # optional, include well-known UEFI certificates into auto-enrollment database (SecureBoot ISO only)
       includeWellKnownCertificates: true
+       # optional, how systemd-boot enrolls SecureBoot keys on first boot: off, manual, if-safe, force
+       # defaults to if-safe (auto-enrolls only in a VM); use force for unattended bare-metal enrollment in setup mode
+      enrollKeys: force
     bootloader: sd-boot # optional, defaults to auto (bootloader chosen by imager), other options: dual-boot, grub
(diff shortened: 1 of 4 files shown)


smira (approved):
🆒

shanduur (approved):

frezbo (approved):
I guess pull in new talos too?

smira:
> I guess pull in new talos too?

I already did previously, nothing new here (it's just an existing option to be wired into the schematic)

PR #483, opened 2026-06-12
refactor: abstract versioned cache

Orzelius:
I didn't wan't to duplicate the code a fourth time for the extra extensions so here's a little abstraction

### DIFF
--- a/internal/cache/single_flight.go
+++ b/internal/cache/single_flight.go
@@ -1,0 +1,73 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package cache
+
+import (
+	"context"
+	"sync"
+
+	"golang.org/x/sync/singleflight"
+)
+
+// SingleFlightCache lazily fetches and caches a list of values keyed by a version tag.
+//
+// Fetches are de-duplicated across concurrent callers via an internal
+// singleflight group, and respect the caller's context for cancellation.
+// Failed fetches are not cached.
+type SingleFlightCache[T any] struct {
+	fetch   func(tag string) (T, error)
+	entries map[string]T
+	sf      singleflight.Group
+	mu      sync.Mutex
+}
+
+// NewSingleFlightCache creates a cache that uses fetch to populate missing entries.
+func NewSingleFlightCache[T any](fetch func(tag string) (T, error)) *SingleFlightCache[T] {
+	return &SingleFlightCache[T]{
+		entries: make(map[string]T),
+		fetch:   fetch,
+	}
+}
+
+// Get returns the cached entry for a tag, fetching it if absent.
+func (c *SingleFlightCache[T]) Get(ctx context.Context, tag string) (T, error) {
+	c.mu.Lock()
+	entry, ok := c.entries[tag]
+	c.mu.Unlock()
+
+	if ok {
+		return entry, nil
+	}
+
+	resultCh := c.sf.DoChan(tag, func() (any, error) { //nolint:contextcheck
+		item, err := c.fetch(tag)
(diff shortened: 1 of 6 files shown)


internal/cache/versioned.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// VersionedCache lazily fetches and caches a list of values keyed by a version tag.
+//
+// Fetches are de-duplicated across concurrent callers via an internal
+// singleflight group, and respect the caller's context for cancellation.
+// Failed fetches are not cached.
+type VersionedCache[T any] struct {
+	fetch   func(tag string) ([]T, error)
+	entries map[string][]T
```
smira: in order to make it more generic, I would replace `[]T` here with `T`, as nothing in this code relies on it being a slice actually.

On error, return zero value of T (`var zero T`) instead of `nil`.

This would make this code work with any `T` type.

internal/cache/versioned.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"golang.org/x/sync/singleflight"
+)
+
+// VersionedCache lazily fetches and caches a list of values keyed by a version tag.
+//
+// Fetches are de-duplicated across concurrent callers via an internal
+// singleflight group, and respect the caller's context for cancellation.
+// Failed fetches are not cached.
+type VersionedCache[T any] struct {
```
smira: Probably something like `SingleflightCache` (as it's not really versioned in any way). It's rather unbounded cache with singleflight protection?

internal/cache/single_flight.go
```diff
@@ -0,0 +1,73 @@
...
+// Failed fetches are not cached.
+type SingleflightCache[T any] struct {
+	fetch   func(tag string) (T, error)
+	entries map[string]T
+	sf      singleflight.Group
+	mu      sync.Mutex
+}
+
+// NewSingleFlightCache creates a cache that uses fetch to populate missing entries.
+func NewSingleFlightCache[T any](fetch func(tag string) (T, error)) *SingleflightCache[T] {
```
smira: either

```suggestion
func NewSingleflightCache[T any](fetch func(tag string) (T, error)) *SingleflightCache[T] {
```

smira: or make them both `SingleFlight`

internal/cache/single_flight_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+func TestVersionedCache(t *testing.T) {
+	t.Parallel()
+
+	t.Run("fetches once and caches the result", func(t *testing.T) {
+		t.Parallel()
+
+		var calls int
+
+		c := cache.NewSingleFlightCache(func(tag string) ([]string, error) {
```
smira:
```suggestion
		c := cache.NewSingleFlightCache(func(tag string) (string, error) {
```

internal/cache/single_flight_test.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+
+	t.Run("propagates fetch errors without caching", func(t *testing.T) {
+		t.Parallel()
+
+		fetchErr := errors.New("boom")
+
+		var calls int
+
+		c := cache.NewSingleFlightCache(func(string) ([]string, error) {
```
smira:
```suggestion
		c := cache.NewSingleFlightCache(func(string) (struct{}, error) {
```

internal/cache/single_flight_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+		var calls int
+
+		c := cache.NewSingleFlightCache(func(tag string) ([]string, error) {
+			calls++
+
+			return []string{tag}, nil
+		})
+
+		got, err := c.Get(context.Background(), "v1.0.0")
```
smira:
```suggestion
		got, err := c.Get(t.Context(), "v1.0.0")
```

smira (approved):

PR #484, opened 2026-06-15
feat: narrow sbom cache key to extension list only

### LINKED ISSUE #478: improve caching of SBOM/scan results
Short version:

1. Derive proper caching key for SBOM/scan result.
2. Do singleflight for both SBOM generation based on the key and for the scan itself.

See also #476 

Cache key - we can try to follow the path established for image building:

1. Schematic is mapped into the imager's Profile. Here SBOM depends only on the extensions list, so don't need to populate anything else.
2. Profile is cleaned up, marshaled and hashed.
3. This provides an SBOM key which already includes schematic, Talos version and arch.
4. Add singleflight based on this key around SBOM generation, use it for caching.
5. For the security scan - same applies, the only new stuff is invalidation based on VEX/vulndb. We should also do singleflight.

Test plan:

1. Get an  SBOM/scan for a schematic A.
2. Mutate schematic in a way that shouldn't affect the SBOM - e.g. test with embedded config, kernel args, etc.
3. Verify that SBOM for schematic B is delivered (almost) immediately - it should be cached.

shanduur:
SBOM content depends only on the extension list, Talos version, and
architecture. Previously the cache key included the full schematicID,
causing cache misses when non-SBOM fields changed (kernel args, config,
meta, secureboot, etc.).

Now hash only the relevant inputs so schematics with identical
extension lists share cached bundles and scan results. Singleflight
uses the same hash for both SPDX build and vulnerability scan.

Fixes https://github.com/siderolabs/image-factory/issues/478

### DIFF
--- a/enterprise/spdx/builder/hash.go
+++ b/enterprise/spdx/builder/hash.go
@@ -1,0 +1,52 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+//go:build enterprise
+
+package builder
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"slices"
+	"sort"
+)
+
+// Hash returns a content hash describing the inputs that determine the SPDX
+// bundle content: the extension list, Talos version, and architecture. It is
+// used as the OCI cache tag, so that:
+//
+//   - Two schematics with the same extension list, version and architecture
+//     share a single cached bundle even when other schematic fields differ.
+//   - Fixes to the SPDX extraction/merge logic can invalidate previously
+//     cached bundles via errata strings (see internal/profile.Hash).
+func Hash(extensions []string, version, arch string) string {
+	hasher := sha256.New()
+
+	// Format version so the hash scheme can be evolved in the future.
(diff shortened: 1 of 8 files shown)


enterprise/spdx/builder/hash.go
```diff
@@ -0,0 +1,51 @@
...
+	hasher := sha256.New()
+
+	// Format version so the hash scheme can be evolved in the future.
+	hasher.Write([]byte("sbom/v1"))
+	hasher.Write([]byte{0})
+
+	// Sort extensions for deterministic hashing regardless of schematic order.
+	sorted := make([]string, len(extensions))
+	copy(sorted, extensions)
+	sort.Strings(sorted)
```
frezbo, line 34: slices.Clone

enterprise/spdx/builder/hash.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for _, ext := range sorted {
+		hasher.Write([]byte(ext))
+		hasher.Write([]byte{0})
+	}
+
+	hasher.Write([]byte(version))
+	hasher.Write([]byte{0})
+	hasher.Write([]byte(arch))
```
frezbo, line 43: missing null byte after arch

frezbo (approved):

PR #491, opened 2026-07-02
feat: add support for extra extensions

### LINKED ISSUE #395: support for "extra/overlaid" extensions/overlays
(Enterprise only feature).

Add support for specifying extra extensions/overlays catalogs which enhance the standard set of extensions/overlays for a specific version of Talos.

Orzelius:
Image factory can now be configured with an extra extensions manifest image allowing to easily add custom extensions in addition to the official Sideo extensions.

closes https://github.com/siderolabs/image-factory/issues/395


cmd/image-factory/cmd/options.go
```diff
@@ -502,6 +517,17 @@ type EnterpriseOptions struct {
...
+	ExtraExtensions ExtraExtensionsOptions `koanf:"extraExtensions"`
+}
+
+// ExtraExtensionsOptions configures custom extensions offered alongside the official ones.
+type ExtraExtensionsOptions struct {
+	// Manifest specifies the OCI repository holding the extra extensions manifest image.
+	//
+	// It may live in a different registry than the official images.
+	Manifest OCIRepositoryOptions `koanf:"manifest"`
 }
```
Orzelius, line 531: I put it under a field just in case we ever want to add another option besides the `Manifest`

cmd/image-factory/cmd/service.go
```diff
@@ -486,7 +486,14 @@ func buildArtifactsManager(logger *zap.Logger, opts Options) (*artifacts.Manager
 		TalosctlImage:          opts.Artifacts.Core.Components.Talosctl,
 
 		ExternalURL: opts.HTTP.ExternalURL,
-	})
+	}
+
+	if enterprise.Enabled() {
+		artifactsManagerOpts.ExtraExtensionManifestImage = opts.Enterprise.ExtraExtensions.Manifest.Image()
+		artifactsManagerOpts.ExtraExtensionsImageRegistry = opts.Enterprise.ExtraExtensions.Manifest.Registry
+	}
```
Orzelius, line 494: This is currently the only gate keeping this enterprise. Technically should be enough but lmk if you think it should be stricter.

smira: I think we should add a reverse failure - if !enterprise.Enabled && this option is not empty - IF should fail to start

smira (approved):

cmd/image-factory/cmd/service.go
Orzelius: Currently non of the other options have the check so I decided against adding it just for extra extensions

PR #492, opened 2026-07-07
feat: add llms.txt for better LLM usage

rothgar:
Adds the file to the root path for automatic discovery and helping LLMs discover endpoints and usage.

### DIFF
--- a/internal/frontend/http/templates/llms.txt
+++ b/internal/frontend/http/templates/llms.txt
@@ -1,0 +1,186 @@
+# Talos Linux Image Factory
+
+The Image Factory builds and serves customized Talos Linux boot artifacts. Artifacts are parameterized by a **schematic** (a hash-identified YAML document describing customizations) and a **Talos Linux version**. The official instance is at `https://factory.talos.dev` (HTTP/OCI) and `https://pxe.talos.dev` (PXE).
+
+## Core Concepts
+
+**Schematic**: A YAML document that describes image customizations. Submitting one returns a stable SHA-256 ID. The same customizations always produce the same ID. The well-known default schematic (no customizations) has ID `376567988ad370138ad8b2698212367b8edcb69b5fd68c80be1f2ec7d603b4ba`.
+
+**Version**: A Talos Linux release version string, e.g. `v1.5.0`. Use `GET /versions` to list available versions.
+
+**Model**: The combination of a schematic ID and a Talos version. From a model you can derive any artifact type.
+
+## Typical Workflow
+
+1. POST a schematic → get back an ID
+2. GET `/versions` to find a version to target
+3. GET `/image/:schematic/:version/:path` to download an artifact
+
+## HTTP API (`https://factory.talos.dev`)
+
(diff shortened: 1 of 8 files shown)


shanduur:
Do we need to write this down? It will become obsolete each time we update docs. If this is needed, maybe we can serve it from Image Factory itself on `factory.talos.dev/llms.txt`?

rothgar:
> Do we need to write this down? It will become obsolete each time we update docs. If this is needed, maybe we can serve it from Image Factory itself on `factory.talos.dev/llms.txt`?

That was the goal and I forgot to also push those changes. I just added a handler for /llms.txt with the text returned.

rothgar:
The lint check failures are not because of this PR.

```
#27 ERROR: process "/bin/sh -c dis-vulncheck -tool=false ./..." did not complete successfully: exit code: 1
------
 > [lint-govulncheck 2/2] RUN --mount=type=cache,target=/root/.cache/go-build,id=image-factory/root/.cache/go-build --mount=type=cache,target=/go/pkg,id=image-factory/go/pkg dis-vulncheck -tool=false ./...:
10.47 
10.47 Vulnerability #2: GO-2026-5932
10.47 The golang.org/x/crypto/openpgp package is unmaintained, unsafe by design, and has known security issues
10.47 More info: https://pkg.go.dev/vuln/GO-2026-5932
10.47 Affected products: 
10.47  - pkg:golang/golang.org%2Fx%2Fcrypto@v0.52.0
10.47 Your code is affected by this vulnerability
10.47 
10.47 
10.47 dis-vulncheck scan complete! 🚀
------
WARNING: No output specified with remote driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:237
--------------------
 235 |     FROM base AS lint-govulncheck
 236 |     WORKDIR /src
 237 | >>> RUN --mount=type=cache,target=/root/.cache/go-build,id=image-factory/root/.cache/go-build --mount=type=cache,target=/go/pkg,id=image-factory/go/pkg dis-vulncheck -tool=false ./...
 238 |     
 239 |     # runs unit-tests with race detector
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c dis-vulncheck -tool=false ./..." did not complete successfully: exit code: 1
make[1]: *** [Makefile:204: target-lint-govulncheck] Error 1
make[1]: Leaving directory '/home/runner/_work/image-factory/image-factory'
make: *** [Makefile:254: lint-govulncheck] Error 2
```

internal/frontend/http/templates/llms.txt
shanduur, line 1: Can we render this on build? This is technically just a docs/API.md with some minor changes, right?

rothgar: Yes, it's just a raw text endpoint. I'm not sure what the difference would be with rendering on build vs embedding. Happy to do it a different way, I just wanted something that was easier for discovery and usage.

shanduur: My thinking is that when we update the `docs/...` pages, we might miss updating the `llms.txt` - by making it  a build-time generated thing (or a part of `make generate` target), we can make sure those are properly synced.

rothgar: The current llms.txt was generated by an LLM. I'm assuming we don't want builds to be dependent on an external LLM but I'm curious if you have ideas on other ways we could generate it from docs.

shanduur: Okay. So let's do this as is. Please move the `llms.txt` to `templates/` and add docs page documenting how it was written and how can we improve it in the future. Then we'll merge it :D

shanduur (approved):

PR #493, opened 2026-07-09
feat: support registry namespace prefix for core artifacts

ruadhan:
## Summary

Only the registry *host* of `artifacts.core` could be overridden, so pull-through caches which prefix the upstream repository path — e.g. a Harbor proxy-cache project — could not be used: the configurable component images worked, but the extension/overlay images discovered inside the manifests were rewritten to the registry host with their upstream path kept as-is, dropping the required prefix.

This adds `artifacts.core.namespace`, an optional repository path prefix prepended to every image pulled from the core registry — the component images as well as the extension and overlay images referenced by the manifests:

```yaml
artifacts:
  core:
    registry: harbor.example.com
    namespace: ghcrio
```

now pulls `ghcr.io/siderolabs/gvisor` via `harbor.example.com/ghcrio/siderolabs/gvisor`.

## Implementation notes

- The prefix is applied to **pull references only**. `ExtensionRef.TaggedReference` (matched against schematic extension names, the UI filter, the meta API, and the official/extra dedupe) is unchanged; the pull reference is carried separately.
- Overlay pull references are rebuilt from the configured registry at fetch time, so they pick up the prefix without touching overlay identity.
- The enterprise extra-extensions registry is unaffected (its manifest is operator-authored, so image paths can be authored with any required prefix).
- `docs/configuration.md` regenerated via docgen (also picks up the previously missing `enterprise.extraExtensions` section).

## Testing

- New unit tests for `extractExtensionList` covering identity vs. pull reference with and without a namespace.
- `helm template` renders the chart correctly; `helm lint` and schema generation (`make chart-gen-schema`) pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/artifacts/fetch.go
+++ b/internal/artifacts/fetch.go
@@ -29,7 +29,7 @@ func (m *Manager) fetchImageByTag(imageName, tag string, architecture Arch, imag
 }
 
 // fetchImageByTag contains combined logic of image handling: heading, downloading, verifying signatures, and exporting.
-func (m *Manager) fetchImageByTagWithRepo(imageName, tag string, reg name.Registry, architecture Arch, imageHandler imagehandler.Handler) error {
+func (m *Manager) fetchImageByTagWithRepo(imageName, tag string, reg registryWithNamespace, architecture Arch, imageHandler imagehandler.Handler) error {
 	// set a timeout for fetching, but don't bind it to any context, as we want fetch operation to finish
 	ctx, cancel := context.WithTimeout(context.Background(), FetchTimeout)
 	defer cancel()
@@ -121,7 +121,7 @@ func (m *Manager) extractOverlay(arch Arch, ref OverlayRef) error {
 
 // fetchExtensionImage fetches a specified extension image and exports it to the storage as OCI.
 func (m *Manager) fetchExtensionImage(arch Arch, ref ExtensionRef, destPath string) error {
-	imageRef := ref.TaggedReference.Digest(ref.Digest)
+	imageRef := ref.pullReference.Digest(ref.Digest)
 
 	if err := m.fetchImageByDigest(imageRef, arch, imagehandler.OCI(destPath+tmpSuffix)); err != nil {
 		return err
--- a/internal/artifacts/manager.go
+++ b/internal/artifacts/manager.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
 	"slices"
 	"sync"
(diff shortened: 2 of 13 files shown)


internal/artifacts/manager.go
```diff
@@ -71,20 +88,25 @@ func NewManager(logger *zap.Logger, options Options) (*Manager, error) {
...
-	imageRegistry, err := name.NewRegistry(options.ImageRegistry, opts...)
+	registry, err := name.NewRegistry(options.ImageRegistry, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse image registry: %w", err)
 	}
 
-	var extraExtensionsImageRegistry name.Registry
+	imageRegistry := registryWithNamespace{
+		registry:  registry,
+		namespace: strings.Trim(options.ImageRegistryNamespace, "/"),
```
shanduur: Don't Trim the `Namespace`, if it is invalid, just reject it in options validation.

internal/artifacts/versions_test.go
```diff
@@ -0,0 +1,116 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+package artifacts
```
shanduur: This should be an `artifacts_test` package.

internal/artifacts/fetch.go
```diff
@@ -121,7 +121,7 @@ func (m *Manager) extractOverlay(arch Arch, ref OverlayRef) error {
 
 // fetchExtensionImage fetches a specified extension image and exports it to the storage as OCI.
 func (m *Manager) fetchExtensionImage(arch Arch, ref ExtensionRef, destPath string) error {
-	imageRef := ref.TaggedReference.Digest(ref.Digest)
+	imageRef := ref.pullReference.Digest(ref.Digest)
 
 	if err := m.fetchImageByDigest(imageRef, arch, imagehandler.OCI(destPath+tmpSuffix)); err != nil {
```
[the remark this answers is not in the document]
ruadhan: Good catch — fixed in 8d0ed9f: `ExtractExtensionSPDX` now pulls via `ref.pullReference` (keeping `TaggedReference.RepositoryStr()` for the SPDX labeling).

internal/artifacts/manager.go
ruadhan: Done in 8d0ed9f — removed the `Trim`; `Options.Validate()` now rejects a namespace whose slash-separated components don't match the distribution reference path-component grammar (go-containerregistry's `name.NewRepository` doesn't reject leading/trailing/double slashes even with strict validation, so it's validated with the grammar regexp directly). Added validation test cases.

internal/artifacts/versions_test.go
ruadhan: Done in 8d0ed9f — moved to `artifacts_test` with an `export_test.go` exposing the internals under test.

internal/artifacts/manager.go
```diff
@@ -30,14 +30,30 @@ import (
...
+// (e.g. a Harbor proxy-cache project) prepended to every repository pulled from it.
+type registryWithNamespace struct {
+	registry  name.Registry
+	namespace string
+}
+
+// Repo returns a repository in the registry with the namespace prefix applied.
+func (r registryWithNamespace) Repo(repoPath string) name.Repository {
+	if r.namespace != "" {
+		repoPath = r.namespace + "/" + repoPath
```
shanduur: This should be done with `path.Join`, then we don’t need to worry about the Validation, as it will handle slashes correctly.

Btw, please don’t post responses copied directly from LLM - if I want to talk to it, I can do it myself. And don’t add Claude as co-author to commits - you are not adding Microsoft as Co-author if you are using VS Code, right?

ruadhan: Sorry, i'm new to go programming and was using Claude to help me, it automaticly responded to the comments. 

Fixed the `path.Join` and removed validation.

shanduur (approved):

shanduur:
Thank you for your contribution!

PR #495, opened 2026-07-15
feat: add audit log for authenticated requests

shanduur:
Records a structured entry per authenticated request (who, what,
when, status, duration, client IP) at the HTTP wrapHandler choke
point. Identity comes from the enterprise auth provider when
present; anonymous requests still record request ID and client IP.

Sink is pluggable via audit.mode: "" (none, default), "file"
(newline-delimited JSON, size-rotated via lumberjack), or "log"
(through the application logger). Immutability is delegated to the
sink's storage (append-only file, WORM bucket, SIEM), not enforced
in-process.


hack/dev/config.yaml
```diff
@@ -37,14 +42,13 @@ authentication:
...
-      registry: registry.local:5000
-      namespace: extension-testing
-      repository: extensions
-      insecure: true
+  # extraExtensions:
+  #   manifest:
+  #     registry: registry.local:5000
+  #     namespace: extension-testing
+  #     repository: extensions
+  #     insecure: true
```
majabojarska, line 51: Is this intended?

shanduur: Yes, this config does not work without any catalog pushed to `registry.local:5000/extension-testing/extensions` - better to leave it commented out IMHO.

smira (approved):

PR #496, opened 2026-07-15
fix: audit file defaults

shanduur:
Set defaults for audit file

### DIFF
--- a/internal/audit/sink/file/file.go
+++ b/internal/audit/sink/file/file.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"sync"
 
 	lumberjack "gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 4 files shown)


internal/audit/sink/file/file.go
```diff
@@ -38,10 +39,10 @@ type Sink struct {
 	mu sync.Mutex
 }
 
-// New creates a file audit sink. A Path of "" writes to stdout without rotation.
+// New creates a file audit sink.
 func New(opts Options) *Sink {
 	if opts.Path == "" {
-		return &Sink{w: os.Stdout}
+		opts.Path = filepath.Join(os.TempDir(), "audit.log")
```
majabojarska, line 48: Nit: Should we move `"audit.log"` to a `const` in `pkg/constants/constants.go`?


```go
const DefaultAuditLogPath = "audit.log"
```

shanduur: It's the only place where it is used, I don't think it is necessary.

majabojarska (approved):

internal/audit/sink/file/file.go
majabojarska: Fair

PR #497, opened 2026-07-15
feat: proxy images through backing registry

### LINKED ISSUE #438: "proxy" source images via Image Factory
This should be an optional feature. The details are TBD.

High-level idea: if someone runs Image Factory on-prem, they need to mirror "source" images, and one of the way is to offer them via an Image Factory itself (e.g. for Enterprise version with private source images). 

So Image Factory offers a mirror capability via internal cache registry, offering a unified view, e.g. `siderolabs/imager` resolves to the underlying `imager` reference, no matter how it is named, and support enough APIs to mirror the whole image (might be even smarter and only list tags/versions matching Talos versions available). This way one could run e.g. `talosctl image-cache` against IF with authentication to mirror all required images.

Other options are also possible, but the idea is to make IF itself a frontend to the source images.

Orzelius:
Allow pulling talos-bundle images through the backing registry. This feature is mainly intented to be used by enterprise customers with a siderolabs hosted image factory configured to provide enterprise versions of sidero images.

closes #438


Orzelius:
some tests took longer than usual, might be because of the cache? Will look into it tomorrow

cmd/image-factory/cmd/options.go
```diff
@@ -543,8 +543,20 @@ type ComponentsOptions struct {
 	// OverlayManifest is the image manifest for overlays.
 	OverlayManifest string `koanf:"overlayManifest"`
 
-	// Talosctl is the image containing the Talos CLI tool.
-	Talosctl string `koanf:"talosctl"`
+	// TalosctlAll is the image containing the Talos CLI tool.
+	TalosctlAll string `koanf:"talosctlAll"`
```
shanduur: This is breaking the config and will require `v1` -> `v2` Image Factory. Let's keep it as `talosctl`.

Orzelius: Does anyone actually use this config yet?

shanduur: We don't know - an on-prem/self-hosted user might use it.

smira: yes, let's please not change the config for no good reason, it's better imho to fix the description/comment to point out it's `talosctl-all` image

hack/release.toml
```diff
@@ -10,6 +10,14 @@ pre_release = false
 
 [notes]
 
+    [notes.image_proxy]
+        title = "Image proxy"
```
smira:
```suggestion
        title = "Source Image Proxy"
```

hack/release.toml
```diff
@@ -10,6 +10,14 @@ pre_release = false
...
+        description = """\
+Allow pulling source images through the backing registry.
```
smira:
```suggestion
Allow pulling source images through the Image Factory.

E.g. `siderolabs/imager` image can be pulled as `factory.talos.dev/siderolabs/imager`
```

hack/release.toml
```diff
@@ -10,6 +10,14 @@ pre_release = false
...
+This feature is mainly intended to be used by enterprise customers with a siderolabs
```
smira: drop these two sentences, wrong capitalization, and they don't really belong to the notes

docs/configuration.md
```diff
@@ -649,12 +649,12 @@ OverlayManifest is the image manifest for overlays.
 
 ---
 
-### `artifacts.core.components.talosctl`
```
smira, line 652: not this file, but new endpoints to mirror images should be documented in the API

internal/frontend/http/registry.go
```diff
@@ -149,20 +154,54 @@ func (f *Frontend) handleExternalRegistry(ctx context.Context, w http.ResponseWr
...
+			out.URL.Scheme = location.Scheme
+			out.URL.Host = location.Host
+			out.URL.Path = location.Path
+			out.URL.RawPath = ""
+
+			if location.RawQuery != "" {
+				out.URL.RawQuery = location.RawQuery
+			}
+			// we don't forward the host header to avoid TLS issues with some registries
+			out.Host = ""
```
smira, line 171: let's drop the Authorization header here as well

internal/frontend/http/registry.go
```diff
@@ -86,57 +88,60 @@ func (img requestedImage) Name() string {
...
 }
 
-func (f *Frontend) handleExternalRegistry(ctx context.Context, w http.ResponseWriter, req *http.Request, imageName, schematicID, manifestsOrBlobs, tagOrDigest string) error {
+// handleImageProxy proxies image requests to the backing registry.
+func (f *Frontend) handleImageProxy(ctx context.Context, w http.ResponseWriter, req *http.Request, route v2Route) error {
+	if f.options.ImageProxy.BackingRegistry.Scheme() != "http" {
+		return xerrors.NewTaggedf[ProxyUnavailableTag]("proxying to an authorized/secure backing registry is not possible")
+	}
+
+	repository, ok := f.options.ImageProxy.Images[route.image]
```
smira, line 116: this is really serious issue - the authentication check (if configured!) is missing on this code path completely!

we should also have a test for it

internal/integration/registry_test.go
```diff
@@ -251,6 +286,93 @@ func testLatestTagResolution(ctx context.Context, t *testing.T, registryAddr str
...
+
+				testInstallerImage(ctx, t, registry, "siderolabs/installer", talosVersion, "", false, platform, baseURL, false, true)
+			})
+		}
+	})
+
+	t.Run("ListTags", func(t *testing.T) {
+		t.Parallel()
+
+		repo := registry.Repo("siderolabs", "installer")
```
smira: a bad path to test, in fact installer is legacy and dropped, either use `installer-base` or `imager`

Orzelius: Thank you, I wasn't aware. Changed it to `installer-base`. I can refactor the rest of the tests in a separate pr as well if that's of any use.

internal/frontend/http/registry.go
Orzelius: I think I'm missing something. [Registering the routes](https://github.com/Orzelius/image-factory/blob/image-proxy/internal/frontend/http/http.go#L183-L184) with the `registerRoute` helper marks them as authenticated. Is there another layer I'm ignorant of?

Orzelius:
@smira, @shanduur Thank you for the feedback, I've made the requested changes.

internal/frontend/http/registry.go
smira: ok, I might have missed it then... let's do negative integration test for Enterprise then (that it doesn't work w/o auth)

Orzelius: Added a basic auth test!

shanduur (approved):

PR #501, opened 2026-07-16
fix: add single-flight around schematic factory

### LINKED ISSUE #465: add singleflight around schematic operations
Add singleflight around schematic get/create by schematic ID.

shanduur:
Fixes #465

### DIFF
--- a/internal/schematic/schematic.go
+++ b/internal/schematic/schematic.go
@@ -8,10 +8,12 @@ package schematic
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/siderolabs/gen/xerrors"
 	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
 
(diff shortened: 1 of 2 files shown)


internal/schematic/schematic.go
```diff
@@ -72,41 +75,71 @@ func (s *Factory) Put(ctx context.Context, cfg *schematic.Schematic) (string, er
...
+	return err
 }
 
 // Get retrieves the stored schematic.
 func (s *Factory) get(ctx context.Context, id string) (*schematic.Schematic, error) {
-	data, err := s.storage.Get(ctx, id)
+	// carry the request ID into the detached call so its logs keep the request_id.
+	reqID := ctxlog.RequestID(ctx)
+
+	sc, err, _ := s.sf.Do("get:"+id, func() (any, error) { //nolint:contextcheck
```
smira: I think we should use same id for both to fix the theoretical issue we had, as it was a race of getting a schematic while put is running

smira (approved):

PR #505, opened 2026-07-22
fix: reuse registry puller for bundle verification

frezbo:
Cosign discovers referrers once, but GetBundles loads each
descriptor by calling ociremote.Bundle. Without
RegistryClientOpts, every bundle load creates a new
go-containerregistry puller and repeats registry discovery and
authentication. A cache entry with 125 historical bundles
therefore performs that setup 125 times.

Pass the current refreshing Image Factory puller through
remote.Reuse for key-based signatures and GSA attestations.
This reuses authentication, transport, and connections while
leaving referrer discovery and cryptographic verification to
cosign.

With the same 125-referrer index, main took 6.46 seconds for
one verification and 25.18 seconds for four. Reusing the
puller took 72 ms and 157 ms respectively.

Signed-off-by: Noel Georgi <git@frezbo.dev>

### DIFF
--- a/enterprise/spdx/storage/registry/registry.go
+++ b/enterprise/spdx/storage/registry/registry.go
@@ -116,7 +116,7 @@ func (s *Storage) Get(ctx context.Context, cacheTag string) (storage.Bundle, err
 	digestRef := s.cacheRepository.Digest(desc.Digest.String())
 
 	// Verify signature
-	err = s.imageSigner.VerifyImage(ctx, digestRef)
+	err = s.imageSigner.VerifyImage(ctx, digestRef, s.puller)
 	if err != nil {
 		ctxlog.Logger(ctx, s.logger).Warn("SPDX bundle signature doesn't validate", zap.Error(err), zap.Stringer("ref", taggedRef))
 
--- a/internal/asset/cache/registry/registry.go
+++ b/internal/asset/cache/registry/registry.go
@@ -87,7 +87,7 @@ func (c *Cache) Get(ctx context.Context, profileID string) (cache.BootAsset, err
 
 	digestRef := c.cacheRepository.Digest(desc.Digest.String())
 
-	err = c.imageSigner.VerifyImage(ctx, digestRef)
+	err = c.imageSigner.VerifyImage(ctx, digestRef, c.puller)
 	if err != nil {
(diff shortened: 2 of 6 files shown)


smira:
Can you provide more details for why it is slow today exactly? I can understand we can cut the time to verify an individual image, but given that it happens once per the lifetime of the IF for a given image, I don't quite understand the benefit.

frezbo:
This has been failing consistently for the https://github.com/siderolabs/talos/actions/runs/29891810188/job/88833571986

especially the talos 1.12 pxe test, it has been failing for a while

frezbo:
So getting `cmdline` from factory took almost a minute

smira:
> This has been failing consistently for the [siderolabs/talos/actions/runs/29891810188/job/88833571986](https://github.com/siderolabs/talos/actions/runs/29891810188/job/88833571986)
> 
> especially the talos 1.12 pxe test, it has been failing for a while

Can you explain to me the failure exactly and why this fixes it please? I understand that getting cmdline would need getting an imager, verifying it, etc., but it anyways does this under singleflight already. 

The CI failure doesn't mean we should diverge from the Cosign code which would be way harder to maintain in the future.

frezbo:
> > This has been failing consistently for the [siderolabs/talos/actions/runs/29891810188/job/88833571986](https://github.com/siderolabs/talos/actions/runs/29891810188/job/88833571986)
> > especially the talos 1.12 pxe test, it has been failing for a while
> 
> Can you explain to me the failure exactly and why this fixes it please? I understand that getting cmdline would need getting an imager, verifying it, etc., but it anyways does this under singleflight already.
> 
> The CI failure doesn't mean we should diverge from the Cosign code which would be way harder to maintain in the future.

hmm, yeh, I went overboard, simple fix would be to just pass in our custom puller so cosign verify is not stuck at 4 jobs

shanduur (approved):

shanduur:
Generally the changes are nice, but maybe we can propose the concurrency control (through allowing to set custom puller or even allow to pass something like MaxConcurrentPulls to the cosign) to the upstream instead of bringing it as a feature ourselves?

shanduur (approved):

smira (approved):

PR #506, opened 2026-07-23
feat: add presigned URL support for image downloads

rsmitty:
I ran into this need with some exploration on image factory enterprise.

This PR adds a new authenticated endpoint which will generate short-lived presigned URLs for image downloads. It works kind of in the way that S3 does with HMAC links that expire after some time.

After generating a schematic ID, a user can post to /presign/image/:schematic/:version/:path with auth and this will return a JSON response with a URL that contains expiry and signature query params. The signature is verified in withAuth before the auth middleware, so presigned downloads need no credentials. Default expiry is 5m.

The HMAC key is auto-generated on startup (zero config) or can be set explicitly via presignedURLKey for multi-replica deployments.

Part of why I needed this is to get away from the need for user/pass when using Omni connected to IFE. With this, Omni can query IFE to generate a presigned URL when a user clicks the download on one of the images they build.

Related to https://github.com/siderolabs/image-factory/issues/498, but not implemented in the same way as described.

### DIFF
--- a/cmd/image-factory/cmd/service.go
+++ b/cmd/image-factory/cmd/service.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"crypto"
 	"crypto/tls"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http"
@@ -45,6 +46,7 @@ import (
 	auditlog "github.com/siderolabs/image-factory/internal/audit/sink/log"
 	frontendhttp "github.com/siderolabs/image-factory/internal/frontend/http"
 	"github.com/siderolabs/image-factory/internal/image/signer"
+	"github.com/siderolabs/image-factory/internal/presign"
 	"github.com/siderolabs/image-factory/internal/remotewrap"
 	"github.com/siderolabs/image-factory/internal/schematic"
 	schematiccache "github.com/siderolabs/image-factory/internal/schematic/storage/cache"
@@ -146,12 +148,18 @@ func RunFactory(ctx context.Context, logger *zap.Logger, opts Options) error {
 
 	defer auditSink.Close() //nolint:errcheck
 
+	presignSigner, err := buildPresignSigner(opts)
+	if err != nil {
+		return err
+	}
+
 	frontendOptions, err := buildFrontendOptions(cacheImageSigner, authProvider, opts)
 	if err != nil {
 		return err
 	}
 
 	frontendOptions.AuditSink = auditSink
+	frontendOptions.PresignedURLSigner = presignSigner
 
 	frontendHTTP, err := frontendhttp.NewFrontend(
 		logger,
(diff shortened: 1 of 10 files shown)


internal/frontend/http/http.go
```diff
@@ -170,10 +172,13 @@ func NewFrontend(
 	registerPublicRoute(frontend.router.GET, "/readyz", frontend.handleReady)
 	registerPublicRoute(frontend.router.HEAD, "/readyz", frontend.handleReady)
 
-	// images - require auth
+	// images - require auth (presigned URLs bypass auth via signature verification)
 	registerRoute(frontend.router.GET, "/image/:schematic/:version/:path", frontend.handleImage)
 	registerRoute(frontend.router.HEAD, "/image/:schematic/:version/:path", frontend.handleImage)
 
+	// presign - auth-protected endpoint that issues presigned URLs
+	registerRoute(frontend.router.POST, "/presign/image/:schematic/:version/:path", frontend.handlePresign)
```
shanduur, line 180: As this is an enterprise-specific feature, it should be defined as an enterprise plugin, similarly to e.g. `spdx` endpoint:
https://github.com/siderolabs/image-factory/tree/main/enterprise

internal/integration/auth_test.go
```diff
@@ -470,6 +477,109 @@ func testAuthCDNNoRedirect(t *testing.T, pool *dockertest.Pool) {
...
+	t.Run("TamperedSignatureRejected", func(t *testing.T) {
+		t.Parallel()
+
+		presignedURL := getPresignedURL(ctx, t, baseURL, schematicID, "v1.9.0", "kernel-amd64")
+
+		// Tamper with the signature.
+		tampered := presignedURL[:len(presignedURL)-1] + "0"
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, tampered, nil)
+		require.NoError(t, err)
```
shanduur, line 533: If the generated signature already ends in 0, this produces the original URL and makes the rejection test fail intermittently.

```suggestion
		replacement := byte('0')
		if presignedURL[len(presignedURL)-1] == replacement {
			replacement = '1'
		}
 		tampered := presignedURL[:len(presignedURL)-1] + string(replacement)
 		
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, tampered, nil)
		require.NoError(t, err)
```

cmd/image-factory/cmd/service.go
```diff
@@ -215,6 +223,30 @@ func buildAuthProvider(logger *zap.Logger, opts Options) (enterprise.AuthProvide
...
+	}
+
+	if key := opts.Authentication.PresignedURLKey; key != "" {
+		decoded, err := base64.StdEncoding.DecodeString(key)
+		if err != nil {
+			return nil, fmt.Errorf("presignedURLKey must be base64-encoded: %w", err)
+		}
+
+		return presign.NewSigner(decoded, ttl), nil
+	}
```
shanduur, line 245: That option is documented as a base64-encoded 32-byte HMAC key, but values like `YQ==` are accepted, resulting in weak signing keys. We should add a `len(decoded) == 32` validation (https://github.com/siderolabs/image-factory/blob/a81f6e9/cmd/image-factory/cmd/options.go#L19) and fail startup otherwise.

rsmitty:
sounds like we may not want this after talking with Andrey. let me come back to your feedback if we still need it.

PR #509, opened 2026-07-24
feat: support JWT download tokens

rsmitty:
This PR allows for identity-scoped JWT download tokens. A token covers all schematics owned by the identity. In the case of htaccess with IFE, that'd be the username. In the future case of auth0, that's the org_id that we'll have for each Omni deployment.

Tokens are short-lived by default at 5m and there is protection to make sure that tokens
 can't be used for anything other than reading.

As it stands in this PR, image factory mints these tokens, but we can chat later if we'd like to have other auth backends do the minting.

How I foresee this working in Omni (auth0 flow):

- Org is added to Auth0 and a machine-to-machine (M2M) app is configured for Omni use.
- Omni is brought up with the M2M key that allows for auth0 authentication.
- User generates installation media.
- When Omni presents the download links, it calls IF to gen a 5m token that's valid for all schematics owned by org_id.
- User clicks download link and something like `https://<if-url>/image/<schematic>/<version>/metal-amd64.iso?token=xxx` is called.

Similar story with current htaccess method, just will bypass the Auth0 setup and will use basic auth against IF to get the token.


rsmitty:
Fun example of this working w/ basic auth in terminal:

```bash
❯ curl http://localhost:8080/.well-known/jwks.json

{"keys":[{"use":"sig","kty":"EC","crv":"P-256","alg":"ES256","x":"bJcKJOLrEX7ex0gJfk3X3r32UOWRYPGiFweNcriIitU","y":"Nzyd-LJgctgR--Rr24KGs-iaK2-k-9_mLO66Lyz0gH4"}]}%

❯ curl -u developer:SideroTest -X POST http://localhost:8080/schematics \
-H "Content-Type: application/yaml" \
-d '
customization:
  systemExtensions:
    officialExtensions:
      - siderolabs/qemu-guest-agent
'
{"id":"979b0fc9393e754bd00e47f59c41b34e2ae502007c5749b8a85fe5811e0a1b65","schematic":"owner: developer\ncustomization:\n    systemExtensions:\n        officialExtensions:\n            - siderolabs/qemu-guest-agent\n"}

❯ TOKEN=$(curl -su developer:SideroTest -X POST http://localhost:8080/download-token | jq -r .token)

❯ echo $TOKEN

eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE3ODQ5MTkyMDAsImlhdCI6MTc4NDkxODkwMCwiaXNzIjoiaW1hZ2UtZmFjdG9yeSIsInN1YiI6ImRldmVsb3BlciJ9.VES6vgEk7P16cpCg0F2KyCdU8uaYlzy8UhEHWKrXJTZWuZZ89QQajMBvt9vWdk64h69SJ2RlULWcn7ZsPVcbYQ


❯ curl -o metal-amd64.iso "http://localhost:8080/image/979b0fc9393e754bd00e47f59c41b34e2ae502007c5749b8a85fe5811e0a1b65/v1.9.4/metal-amd64.iso?token=$TOKEN"

  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed
100 101.2M 100 101.2M   0      0 52.48M      0   00:01   00:01         52.49M
```

internal/downloadtoken/downloadtoken.go
```diff
@@ -0,0 +1,148 @@
...
+func NewIssuer(privateKey *ecdsa.PrivateKey, ttl time.Duration) (*Issuer, error) {
+	sig, err := jose.NewSigner(
+		jose.SigningKey{Algorithm: jose.ES256, Key: privateKey},
+		(&jose.SignerOptions{}).WithType("JWT"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("downloadtoken: failed to create signer: %w", err)
+	}
+
+	pubJWK := jose.JSONWebKey{
```
shanduur: Could we assign a stable `kid` to this JWK and put the same value in the JWT protected header? The current single-key setup works, but consumers cannot select the correct key once rotation introduces multiple keys. A configured key ID or a deterministic public-key thumbprint would both work.

internal/frontend/http/http.go
```diff
@@ -318,11 +319,29 @@ func (f *Frontend) withAuth(h Handler, requireAuth bool, username *string) Handl
...
-	return authProvider.Middleware(func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error {
-		*username, _ = authProvider.UsernameFromContext(ctx)
+	return func(ctx context.Context, w http.ResponseWriter, r *http.Request, p httprouter.Params) error {
+		// Download token: if the request carries a valid JWT on a read-only
+		// method, extract the subject as the authenticated identity. Ownership
+		// is enforced normally by schematicFactory.Get() since the JWT subject
+		// is set on the context. Tokens are only accepted on GET/HEAD to
+		// prevent use on write endpoints (POST /schematics, etc.).
+		if f.options.DownloadTokenIssuer != nil && (r.Method == http.MethodGet || r.Method == http.MethodHead) && r.URL.RawQuery != "" {
+			if tokenStr := r.URL.Query().Get("token"); tokenStr != "" {
```
shanduur, line 330: Could we accept this through `Authorization: Bearer <token>` instead of the query string? URLs commonly appear in access and proxy logs, browser history, monitoring data, and referrer headers. If signed URLs are an intentional product requirement, the URL itself should be documented as a secret and token query parameters should be redacted throughout the serving path.

internal/downloadtoken/downloadtoken.go
```diff
@@ -0,0 +1,148 @@
...
+// Issue creates a signed JWT for the given subject (org_id or username).
+func (i *Issuer) Issue(subject string) (string, error) {
+	now := time.Now()
+
+	claims := jwt.Claims{
+		Subject:  subject,
+		Issuer:   issuerName,
+		IssuedAt: jwt.NewNumericDate(now),
+		Expiry:   jwt.NewNumericDate(now.Add(i.ttl)),
+	}
```
shanduur, line 122: Could we bind the token to its intended purpose in both issuance and verification? At minimum, add an audience such as `image-factory-download` and a scope or purpose such as `image:download`. If one token does not need to cover every schematic owned by the identity, also consider binding it to the schematic, version, and asset. Without purpose claims, any JWT signed by this key with matching issuer, subject, and time claims can be treated as a download token.

enterprise/downloadtoken/frontend.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	token, err := f.issuer.Issue(username)
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
+	return json.NewEncoder(w).Encode(map[string]string{
```
shanduur: Would it make sense to return the conventional bearer-token shape here: `access_token`, `token_type: "Bearer"`, and `expires_in`? It makes the transport contract and lifetime explicit, and is more interoperable than a custom `token` field without expiry metadata.

shanduur (commented):
Few comments, but overall LGTM.

enterprise/downloadtoken/frontend.go
rsmitty: fixed in amended commit. token now looks like:

```bash
❯ curl -su developer:SideroTest -X POST http://localhost:8080/download-token | jq
{
  "access_token": "eyJhbGciOiJFUzI1NiIsImtpZCI6IlJCQ0NMYWFrZzNSbTNIX2N0b3RKSEZfMlZWeF9BdUk5STZ3bU5RMGdVZGciLCJ0eXAiOiJKV1QifQ.eyJhdWQiOiJpbWFnZS1mYWN0b3J5OmRvd25sb2FkIiwiZXhwIjoxNzg0OTIzOTI0LCJpYXQiOjE3ODQ5MjM2MjQsImlzcyI6ImltYWdlLWZhY3RvcnkiLCJzdWIiOiJkZXZlbG9wZXIifQ.pPeLma6z1Q_i1nG7CCPUm7zZ6qI3gp0RgWeQLoqBSckoC3kS2PB0QPbHMryhQH3QzDpcGMaC5ixrIfq8TVdDAg",
  "token_type": "Bearer",
  "expires_in": 300
}
```

internal/downloadtoken/downloadtoken.go
```diff
@@ -0,0 +1,148 @@
...
+	var claims jwt.Claims
+
+	if err = tok.Claims(i.key, &claims); err != nil {
+		return "", fmt.Errorf("downloadtoken: failed to verify token: %w", err)
+	}
+
+	if err = claims.ValidateWithLeeway(jwt.Expected{
+		Issuer: issuerName,
+		Time:   time.Now(),
+	}, 0); err != nil {
```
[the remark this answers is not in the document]
rsmitty: fixed in amended commit

internal/frontend/http/http.go
```diff
@@ -318,11 +319,29 @@ func (f *Frontend) withAuth(h Handler, requireAuth bool, username *string) Handl
...
+				if sub, err := f.options.DownloadTokenIssuer.Verify(tokenStr); err == nil {
+					*username = sub
+					ctx = authProvider.ContextWithUsername(ctx, sub)
+
+					return h(ctx, w, r, p)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
rsmitty: fixed in amended commit

internal/downloadtoken/downloadtoken_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// TTL of 0 means the token expires immediately.
+	issuer, err := downloadtoken.GenerateIssuer(0)
+	require.NoError(t, err)
+
+	token, err := issuer.Issue("org_abc123")
+	require.NoError(t, err)
+
+	// Sleep briefly to ensure expiry.
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
rsmitty: fixed in amended commit.

internal/downloadtoken/downloadtoken.go
rsmitty: added. we now compute the thumbprint of the public key and set it as the kid on both the JWK and JWT header.

internal/downloadtoken/downloadtoken.go
rsmitty: added "image-factory:download" as the audience for both issuance and verification. however, one token will cover all schematics owned by the identity so the rest of this doesn't quite apply.

internal/frontend/http/http.go
rsmitty: my understanding is that we can't accept this through headers. or at least we'd have to support both. I agree that this has a chance to leak the token in logs and stuff, but you can't attach the header in a browser-based download scenario and omni will need to do that.

but I also feel like this is pretty low risk since tokens are only valid for 5m.

internal/integration/auth_test.go
```diff
@@ -362,6 +370,199 @@ func testOwnership(ctx context.Context, t *testing.T, baseURL string) {
...
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
+		require.NoError(t, err)
+
+		resp, err := http.DefaultClient.Do(req)
+		require.NoError(t, err)
+
+		t.Cleanup(func() { resp.Body.Close() }) //nolint:errcheck
+
+		assert.NotEqual(t, http.StatusUnauthorized, resp.StatusCode)
+		assert.NotEqual(t, http.StatusForbidden, resp.StatusCode)
```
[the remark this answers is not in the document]
rsmitty: fixed

enterprise/downloadtoken/frontend.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil
+	}
+
+	token, err := f.issuer.Issue(username)
+	if err != nil {
+		return err
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
```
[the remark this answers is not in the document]
rsmitty: fixed

cmd/image-factory/cmd/service.go
```diff
@@ -215,6 +221,25 @@ func buildAuthProvider(logger *zap.Logger, opts Options) (enterprise.AuthProvide
...
+	if !opts.Authentication.Enabled {
+		return nil, nil //nolint:nilnil
+	}
+
+	ttl := opts.Authentication.DownloadTokenTTL
+	if ttl == 0 {
+		ttl = defaultDownloadTokenTTL
+	}
+
+	return enterprise.NewDownloadTokenIssuer(opts.Authentication.DownloadTokenKeyPath, ttl)
```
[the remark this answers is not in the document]
rsmitty: fixed

internal/downloadtoken/downloadtoken.go
```diff
@@ -0,0 +1,168 @@
...
+
+	block, _ := pem.Decode(data)
+	if block == nil {
+		return nil, fmt.Errorf("downloadtoken: no PEM block found in %s", path)
+	}
+
+	key, err := x509.ParseECPrivateKey(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("downloadtoken: failed to parse EC private key: %w", err)
+	}
```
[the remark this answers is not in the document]
rsmitty: fixed

shanduur (approved):
LGTM. Can you run `make docs` before merging?

PR #511, opened 2026-07-27
feat: add auth0 authentication provider

rsmitty:
This PR adds Auth0 as an authentication provider for enterprise builds, alongside the existing htpasswd provider. Selected via `authentication.provider`, which is now validated at startup.

Setting `domain` and `audience` enables validation of bearer tokens issued directly by Auth0. The tenant's JWKS is resolved through OIDC discovery and cached with a background refresh. Tokens must carry an `org_id` claim, which becomes the identity principal and is stored on the context using the same key htpasswd uses, so ownership checks, audit logging and download tokens work unchanged against Auth0 orgs. Credentials come from `Authorization: Bearer`, falling back to the Basic auth password so Talos and other OCI clients can present a token through the normal registry auth flow.

Browser login is layered on top of this and considered optional. We'll be using both, but feels like there might be a path someday where we (or a self-hosted user) wants to just allow M2M tokens but no UI auth. Setting clientID, clientSecret, redirectURL and sessionKey will enable the browser and then serves /login, /callback and /logout via OAuth2 authorization code + PKCE.

Auth failures are content-negotiated: browsers are redirected to /login, everything else gets a 401 with a Bearer challenge.


rsmitty:
example of using curl to get a token and upload an empty schematic:

```
❯ TOKEN=$(curl -s -X POST https://image-factory-enterprise-dev.us.auth0.com/oauth/token \
    -d grant_type=client_credentials \
    -d client_id=xxx -d client_secret=yyy \
    -d audience=https://image-factory-enterprise-dev -d organization=org_xxx | jq -r .access_token)

❯ echo $TOKEN | cut -d. -f2 | base64 -d 2>/dev/null | jq '{aud, iss, org_id}'
{
  "aud": "https://image-factory-enterprise-dev",
  "iss": "https://image-factory-enterprise-dev.us.auth0.com/",
  "org_id": "org_xxx"
}

❯ curl -H "Authorization: Bearer $TOKEN" http://localhost:8080/schematics -d '{}'
{"id":"1919c646c96f648b5578c94eedec8602089e0a16e51ae7dbb0796a01e399dbf6","schematic":"owner: org_xxx\ncustomization: {}\n"}
```

smira (commented):
1. Open redirect in the browser login flow (security)

enterprise/auth/auth0/browser.go:50 takes return_to straight from the query string, and :190 redirects to it verbatim after a successful login:

returnTo := r.URL.Query().Get("return_to")   // unvalidated
...
http.Redirect(w, r, sc.ReturnTo, http.StatusFound)

The state cookie is AEAD-encrypted so it can't be tampered with, but that's not the attack — the attacker simply sends the victim to https://factory.example.com/login?return_to=https://evil.example/. The victim authenticates against the real factory and is bounced off-site, on a URL that legitimately belongs to your domain. The middleware's own callers always pass a relative path (url.QueryEscape(r.URL.RequestURI())), so restricting this costs nothing:

if returnTo == "" || !strings.HasPrefix(returnTo, "/") || strings.HasPrefix(returnTo, "//") || strings.HasPrefix(returnTo, "/\\") {
    returnTo = "/"
}

Worth validating on read in CallbackHandler too, so an older/rotated cookie can't carry a stale absolute URL.

2. offline_access is never requested — the entire refresh path is dead

browser.go:79 requests "scope": {"openid"}. Auth0 only issues a refresh token when offline_access is in the scope (and the API has refresh tokens enabled). So tokenSet.RefreshToken will always be empty, which means:

- sessionPayload.RefreshToken is always ""
- sessionToken's proactive-refresh branch (provider.go:340, "within 5 minutes of expiry") never fires
- the expired branch always hits errors.New("session expired and no refresh token available")
- doRefresh (browser.go:222) is unreachable, as is its no-rotation fallback

Functionally the user still recovers — cookie cleared, redirect to /login, Auth0 SSO session re-auths silently — so this isn't broken, just misleading. Pick one: add offline_access to the scope and keep the refresh machinery, or drop the refresh code and the RefreshToken field. Right now the code claims a capability it doesn't have, and the 24h cookie lifetime (session.go:92, "long enough for a transparent refresh") is justified by a refresh that can't happen.

3. /callback is hardcoded but redirectURL is free-form config

The route is registered at the literal /callback (http.go:229) and the state cookie is scoped to Path: "/callback" (session.go:134), but redirectURL is an arbitrary operator-supplied URL sent to Auth0 as redirect_uri. Configure https://factory.example.com/auth/callback and Auth0 will faithfully redirect there, into a 404 — with no error at startup and no hint as to why. NewProvider already parses config; parse RedirectURL too and reject anything whose path isn't /callback (or, better, derive the route path from it).

4. Auth failures that redirect produce a bogus audit record

When the middleware redirects instead of returning an error (provider.go:267 and :290), it returns nil. wrapHandler then records status: 200, username: "" in the audit sink for what was an unauthenticated request to a protected endpoint. Every browser hit to / while logged out lands in the audit log as a successful anonymous request. Either return a tagged error and let the error path own the redirect, or plumb the real status through.

5. Startup hard-fails if Auth0 is unreachable

Run (provider.go:223) returns the error from the initial KeyFunc, which fails the errgroup and takes the process down. That's a meaningful behavior change from the htpasswd provider, whose Run is a resilient reload loop. A transient DNS/network blip during a rolling restart will crash-loop the factory even though the JWKS provider would happily fetch lazily on the first request. Consider logging the pre-warm failure and continuing, or retrying with backoff. If fail-fast is the deliberate choice, say so in the comment — it reads as an optimization right now, not a hard dependency.

6. Smaller correctness/robustness points

- No nonce in the authorize request, and oauth.IDTokenValidationOptions{} is empty so nonce validation is skipped. PKCE covers the main injection risk, so this is defense-in-depth rather than a hole — but it's cheap to add alongside the state nonce you already generate.
- VerifyCredentials uses context.Background() (provider.go:310) — if the JWKS cache is cold or refreshing, that call has no deadline and no cancellation. The interface signature doesn't take a ctx, so at minimum wrap it in a context.WithTimeout.
- isSecure depends on X-Forwarded-Proto (browser.go:261). Behind a TLS-terminating proxy that doesn't set that header, session cookies ship without Secure. Since HTTP.ExternalURL is validated as a required absolute URL, using its scheme as a floor (https → always Secure) would be more reliable than trusting a hop header.
- Dead branch: LogoutHandler's returnTo = "/" fallback (browser.go:204) is unreachable — Options.Validate requires externalURL, and service.go always passes it. Also, / wouldn't be a valid Auth0 returnTo anyway (Auth0 requires an absolute, allow-listed URL), so the fallback would fail if it ever were reached.

Test coverage

Bearer-token validation is well covered — the table in provider_test.go hits expiry, audience, issuer, signing key, malformed tokens, and missing org_id, and the integration test covers the Basic-auth-password path and cross-org 403. That's the security-critical surface and it's solid.

The gap is the browser flow, which is where the interesting bugs live and which has essentially zero coverage:

- session.go is entirely untested — no round-trip test for encryptBytes/decryptBytes, no test that a truncated or tampered ciphertext is rejected, no test that a cookie encrypted under key A fails under key B.
- No handler tests for /login, /callback, /logout — state mismatch, missing code, Auth0 error param, and the return_to handling above are all untested. Issues 1–3 would each have been caught by a handler-level test.
- testAuth0BrowserRedirect only asserts 404-when-disabled and 401-for-browsers-without-login; it never exercises browser login e
                                                                                                                                                                                                                                                                                              Also, the "bearer takes precedence over basic auth" case (provider_test.go) is a no-op — it only sets the Bearer header, and itstesting precedence. Since SetBasicAuth and the Bearer set collide on the same header, precedence can't be tested this way. Eitherdrop the case or restructure extractToken to take the two credential sources so the ordering is testable.

cmd/image-factory/cmd/options.go
```diff
@@ -565,19 +563,48 @@ func (c ComponentsOptions) ImageMap() map[string]string {
...
+	//
+	// Domain and audience are always required, and always enable validation of
+	// bearer tokens issued directly by Auth0 (the machine-to-machine path used by
+	// Omni and Talos node image pulls).
+	//
+	// Browser login is an optional addition on top: set clientID, clientSecret,
+	// redirectURL and sessionKey to also serve /login, /callback and /logout for
+	// interactive users.
+	// Machine-to-machine bearer tokens keep working unchanged when it is enabled.
+	// Setting only some of the four is rejected at startup.
```
shanduur: If field is required or optional should be documented directly on the field.

pkg/enterprise/enterprise.go
```diff
@@ -162,3 +155,59 @@ type AuthProvider interface {
 	// JWT subject so that downstream ownership checks work normally.
 	ContextWithUsername(ctx context.Context, username string) context.Context
 }
+
+// Auth0Config holds configuration for the Auth0 authentication provider.
+// Defined here (not in cmd/) to avoid import cycles: pkg/enterprise imports
+// enterprise/auth/auth0, so auth0 cannot import pkg/enterprise.
```
shanduur: nit: I think this is excessive comment - might become out of date. Leave only first line here.

internal/frontend/http/http.go
```diff
@@ -271,7 +275,11 @@ func (f *Frontend) wrapHandler(h Handler, requireAuth bool) httprouter.Handle {
 
 		level, status := MatchError(err, func(message string, code int) {
 			if code == http.StatusUnauthorized {
-				w.Header().Set("WWW-Authenticate", `Basic realm="Image Factory Enterprise"`)
+				// Only set the fallback challenge header if the handler didn't already
+				// set one (e.g. the Auth0 middleware sets Bearer realm=...).
+				if w.Header().Get("WWW-Authenticate") == "" {
+					w.Header().Set("WWW-Authenticate", `Basic realm="Image Factory Enterprise"`)
```
shanduur: nit:

```suggestion
					w.Header().Set("WWW-Authenticate", `Basic realm="Image Factory Enterprise", charset="UTF-8"`)
```

enterprise/auth/auth0/provider.go
```diff
@@ -0,0 +1,495 @@
...
+			authentication.WithClientID(cfg.ClientID),
+			authentication.WithClientSecret(cfg.ClientSecret),
+		}
+
+		if cfg.HTTPClient != nil {
+			authOpts = append(authOpts, authentication.WithClient(cfg.HTTPClient))
+		}
+
+		authClient, authErr := authentication.New(ctx, cfg.Domain, authOpts...)
+		if authErr != nil {
```
shanduur: Won't this prevent starting Image Factory all-together in case oof network blip, or transient Auth0/JWKS outage?

enterprise/auth/auth0/provider.go
```diff
@@ -0,0 +1,495 @@
...
+			}
+		}
+
+		if tokenStr == "" {
+			if p.BrowserLoginEnabled() && isBrowserRequest(r) {
+				return p.redirectToLogin(w, r, "no token provided")
+			}
+
+			logger.Debug("auth0: authentication required: no token provided")
+			w.Header().Set("WWW-Authenticate", `Bearer realm="Image Factory Enterprise"`)
```
shanduur: Unauthenticated/non-browser requests set only `WWW-Authenticate: Bearer`, soo the HTTP wrapper then skips adding the Basic fallback because the header is already set, right? 
But this provider explicitly supports Auth0 JWTs in the Basic password for OCI/Talos clients.
Challenge-driven registry clients that wait for a Basic challenge can fail before sending their configured Basic password. 

Maybe we can advertise Basic as well, at least on /v2 registry paths?

PR #512, opened 2026-07-28
feat: serve detached signatures for boot assets

### LINKED ISSUE #13881: add detached signatures for download assets (enterprise)
(See https://github.com/siderolabs/enterprise/issues/11)

Add a detached signature to non-container assets exactly like checksums.

`cosign verify-blob` should pass e.g. for the `.iso` and `.iso.sig`.

shanduur:
Adds a `.sigstore.json` sidecar endpoint alongside the existing
checksum suffixes so clients can verify downloaded assets without
trusting the registry signature path. Bundles are signed on first
request (singleflight-guarded) and cached in the asset cache, keyed
by asset hash + signer identity so key rotation invalidates stale
bundles automatically.

Both KeySigner and GSASigner gain SignBlob/BlobSigningIdentity to
produce message-signature bundles instead of image-attached
signatures, since detached assets have no OCI manifest to attach to.

Closes https://github.com/siderolabs/talos/issues/13881


frezbo:
why not keyless too, it supports it AFAIK

shanduur:
Keyless can’t produce legacy signature and needs to write to a rekor chain. I’m working on that now.

frezbo:
> Keyless can’t produce legacy signature and needs to write to a rekor chain. I’m working on that now.

i think we should not care about legacy signatures at all, cosign is going to v4 soon and it's going to be super legacy then, just only new bundle formats

smira (approved):

smira:
Probably misses `Fixes:` and estimation?

PR #513, opened 2026-07-29
feat(enterprise): publish installer build evidence

### LINKED ISSUE #13882: add provenance/SBOM to the installer image (enterprise)
(See https://github.com/siderolabs/enterprise/issues/11)

Add two more manifests to the `installer` manifest (it's already a multi-arch image):

* SBOM
* provenance (Talos version, schematic ID, IF version)

We could look into how buildkit does that as an example.

`cosign verify-attestation` should be the test case.

shanduur:
Generated Installer images now carry a signed evidence graph:
per-platform SPDX 2.3 SBOM attestations on each manifest and an
index-level SLSA Provenance v1 attestation, both as Sigstore bundles
over OCI referrers. Downstream consumers can verify what went into a
schematic without trusting the Factory serving it.

Publication is mandatory for Talos 1.13.0+ stable releases and has no
runtime toggle: an optional contract cannot be relied on by verifiers.
Talos 1.13 prereleases stay outside the contract.

Referrers fall back to the OCI referrers-tag schema when the backing
registry lacks the native OCI 1.1 API, so registry:2 and registry:3
work unchanged.

Closes https://github.com/siderolabs/talos/issues/13882


internal/integration/registry_test.go
```diff
@@ -266,6 +289,51 @@ func assertImageSignature(ctx context.Context, t *testing.T, ref name.Reference,
...
+	publicKeyPath := filepath.Join(t.TempDir(), "cosign.pub")
+	require.NoError(t, os.WriteFile(publicKeyPath, publicKey, 0o600))
+
+	for _, expected := range []struct {
+		name          string
+		predicateType string
+		reference     name.Digest
+	}{
+		{name: "platform SPDX", predicateType: "https://spdx.dev/Document", reference: platformRef},
+		{name: "index SLSA provenance", predicateType: "https://slsa.dev/provenance/v1", reference: indexRef},
```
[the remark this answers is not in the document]
shanduur: Fixed in 800d428: the integration uses attestation.SPDXPredicateType (https://spdx.dev/Document/v2.3), and the stock Cosign invocation now receives the Enterprise registry credentials.

docs/attestations/installer-build-v1.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+Verify a platform SBOM using that platform manifest's digest:
+
+```shell
+cosign verify-attestation \
+  --key signing-key.pub \
+  --type https://spdx.dev/Document \
+  --insecure-ignore-tlog \
+  factory.example.com/metal-installer/<schematic>@sha256:<platform-digest>
```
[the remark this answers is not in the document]
shanduur: Fixed in 800d428: the documented Cosign command now uses https://spdx.dev/Document/v2.3.

internal/installer/provenance.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	architectures := make([]any, 0, len(input.Platforms))
+	for _, platform := range input.Platforms {
+		architectures = append(architectures, platform.Platform.Architecture)
+	}
+
+	internalParameters, err := structpb.NewStruct(map[string]any{
+		"architectures": architectures,
+	})
```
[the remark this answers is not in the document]
shanduur: Fixed in 800d428: architecture names are sorted before provenance serialization, with the test input reversed to verify deterministic output.

docs/api.md
```diff
@@ -376,6 +376,16 @@ Examples:
 Pulls the Talos Linux `installer` image with the specified schematic and Talos Linux version.
 The image platform (architecture) will be determined by the architecture of the Talos Linux Linux machine.
 
+Enterprise Installer images also publish per-platform SPDX SBOM attestations and index-level SLSA provenance through OCI referrers.
+Image Factory uses the native OCI 1.1 API when available and the OCI referrers-tag schema for registries such as `registry:2` and `registry:3`.
+See [Enterprise Installer Build Evidence v1](attestations/installer-build-v1.md) for the evidence contract and verification commands.
+
+#### `GET /v2/<installer>/<schematic>/referrers/<digest>`
```
smira, line 384: that's a cool idea!

smira (commented):
go.mod is not tidy — CI will fail. internal/image/signer/signer.go:2583 now imports github.com/secure-systems-lab/go-securesystemslib/dsse directly, but the dependency is still // indirect.

enterprise/spdx/builder/builder.go
```diff
@@ -237,8 +247,41 @@ func (b *Builder) extractTalosSPDX(ctx context.Context, bundle *Bundle, versionT
...
+	bufferedReader := bufio.NewReader(reader)
+
+	magic, err := bufferedReader.Peek(4)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to read initramfs compression header: %w", err)
+	}
+
+	switch {
+	case bytes.Equal(magic, xzHeaderMagic):
+		xzReader, err := xz.NewReader(bufferedReader)
```
smira: I don't think we need it in real life - xz is really old Talos, predates 1.13+ Talos a lot.

internal/artifacts/dependency.go
```diff
@@ -0,0 +1,115 @@
...
+	"fmt"
+
+	"github.com/blang/semver/v4"
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+)
+
+// ImageDependency identifies the exact platform manifest consumed from an OCI image.
+type ImageDependency struct {
```
smira, line 18: Provenance dependencies are resolved after the build, independently of it. r

esolveInstallerDependencies (internal/frontend/http/registry.go) re-walks GetOfficialExtensions / GetOfficialOverlays / GetInstallerDependency from scratch, duplicating what profile.EnhanceFromSchematic already did during assetBuilder.Build. Nothing ties the two resolutions together — if the artifacts manager's version/extension cache refreshes between the build and this call, the provenance attests to digests that were not the ones consumed. For a provenance feature that is the central correctness property. findOfficialExtension is also a verbatim copy of the unexported profile.findExtension (internal/profile/profile.go:283), so the two will drift.

I would suggest having EnhanceFromSchematic (or the asset builder) return the resolved ExtensionRef/OverlayRef/base-installer refs it actually used, and feed those into EvidenceInput. 

I would even argue we should not provide much evidence beyond the schematic itself, as this is our cache key basically.

internal/image/signer/gsa_blob_test.go
smira, line 1: Unexplained deletion of internal/image/signer/gsa_blob_test.go (126 lines, 4 tests, added one commit ago in ae757b5), plus TestKeySignerImplementsBlobSigner.

smira (commented):
Hot-path cost on cache hits. verifyCachedInstallerEvidence runs on every manifest request that finds an existing image: a full index Get, then three cosign.VerifyImageAttestations calls (2× SPDX on platform manifests, 1× SLSA on the index), each doing a referrers lookup + bundle fetch + signature verification. That's ~7 extra registry round-trips per installer pull where there was previously one VerifyImage. 

I would say this seems extremely expensive while I don't see why would we do that - if we want to re-generate the installers (existing ones) we could do that with other changes.

But as long as the installer was built, I don't think we should anything beyond regular cache check to serve it - as long as the overarching manifest is signed, we should serve it without looking inside

internal/image/signer/signer.go
```diff
@@ -160,6 +173,98 @@ func (s *KeySigner) BlobSigningIdentity() string {
...
+	if err != nil {
+		return fmt.Errorf("failed to marshal attestation bundle: %w", err)
+	}
+
+	remoteOpts, err := pusher.RemoteOptions()
+	if err != nil {
+		return fmt.Errorf("failed to get remote options for push: %w", err)
+	}
+
+	if err = cosignremote.WriteAttestationNewBundleFormat(
```
smira: Unbounded referrer accumulation on rebuild. 

The SLSA predicate embeds startedOn/finishedOn/invocationId, so a rebuild of an identical index produces a different bundle digest and WriteAttestationNewBundleFormat appends it as an additional referrer to the same subject. Each rebuild permanently adds one more provenance bundle that all future verifications must fetch and check — this compounds (4). SPDX bundles are content-identical and will dedupe; provenance will not. Either make the provenance bundle deterministic for an identical index, or prune/replace existing referrers before publishing.

internal/frontend/http/registry.go
```diff
@@ -138,6 +225,7 @@ func (f *Frontend) handleExternalRegistry(ctx context.Context, w http.ResponseWr
 	}
 
 	location := craftRedirectURL(repo, imageName, schematicID, manifestsOrBlobs, tagOrDigest)
+	location.RawQuery = req.URL.RawQuery
```
smira: is this correct? do we want to forward any query this way?

internal/image/signer/gsa_blob_test.go
shanduur: `//nolint:testpackage` - better to test it as a black box. `TestKeySignerImplementsBlobSigner` does not bring any value, we are testing it on build time.

internal/frontend/http/registry.go
shanduur: Hmm, I thought we might need to for referrers API, but I might be wrong. Let me fix that and test.

internal/image/signer/signer.go
shanduur: Should be fixed now I guess.

enterprise/spdx/builder/builder.go
shanduur: Correct for a 1.13+ evidence support floor, but removing it alone
currently breaks the Enterprise integration matrix, which still builds and
attests versions as old as `v1.3.7`. XZ support was added after that exact CI path
fed an XZ initramfs to the SPDX builder.

smira: We don't need to test going that back for Enteprise imho. We would have dead code just for the test, I would rather eventually drop old versions eventually.

docs/api.md
shanduur: Thanks!

internal/artifacts/dependency.go
shanduur: Good catch. Normal cache hits no longer add referrers. We create the index signature only after publishing and verifying all required evidence. Later requests verify that signature and reuse the immutable index, so they do not rebuild it or publish the evidence again.

I considered deduplicating by subject and predicate type, but that is too weak. It proves that an attestation has the expected subject, type, and signer, but not that its materials and invocation parameters match this build. Reusing the first match could preserve stale provenance. Deterministic predicate bytes would not be enough either, since Sigstore bundles and artifact manifests can still differ between signing runs.

A forced rebuild of the same index can still add another provenance referrer. Avoiding that needs an explicit retention policy, especially since OCI referrers have no replace-by-predicate operation. I would handle that separately rather than quietly adopting "first matching attestation wins" here.

smira (commented):
2. Duplicate overlay material in provenance. The seen dedup map lives on dependencyCapturingProducer, which is constructed per EnhanceFromSchematicWithDependencies call — i.e. per arch — but buildInstallImage concatenates both arches' results. EnhanceFromSchematic resolves overlays twice, for runtime.GOARCH (native, profile.go:560) and prof.Arch (:565), so the host-arch overlay is captured in both iterations. Confirmed empirically with an overlay schematic on an amd64 host:

base-installer|amd64|…installer-base@sha256:aaa…
overlay|amd64|…sbc-rpi-amd64@sha256:ddd…      ← from arch=amd64 pass
base-installer|arm64|…installer-base@sha256:aaa…
overlay|amd64|…sbc-rpi-amd64@sha256:ddd…      ← duplicate, from arch=arm64 pass (native)
overlay|arm64|…sbc-rpi-arm64@sha256:ddd…

Two byte-identical resolvedDependencies entries (same name, uri, digest) land in the SLSA predicate. Move the dedup up to buildInstallImage (or de-dup the concatenated dependencies slice before building EvidenceInput). Related: because the native overlay is host-dependent, the provenance for a given schematic differs by builder host arch — defensible (it genuinely is a build tool) but worth a line in the doc.

3. Input.BaseInstaller.ImageRef is silently rewritten to a digest ref (profile.go:371-377), overriding the value EnhanceFromSchematic deliberately sets and comments as // fake reference. I traced the consequences:
- The imager only reads ImageRef at outInstaller's final tarball.WriteToFile(path, ref, …); ggcr skips RepoTags for non-name.Tag refs, so the intermediate tarball loses its RepoTags. image-factory reads it back with tarball.Image(r, nil), so the published installer index digest is unaffected. No breakage — but the effect is undocumented and easy to trip over later.
- The real effect is on profile.Hash → internal/asset/asset.go:144, the asset-build cache key. TestEnhanceFromSchematicCapturesConsumedDependencies asserts this (require.NotEqual(firstHash, secondHash)), so it's clearly deliberate: the asset cache now tracks the base-installer digest instead of just the tag. That's a genuine cache-correctness improvement — but it's a behavior change to asset caching buried in an attestation PR, with no mention in the commit message or docs, and it invalidates existing installer asset cache entries once on deploy. Please call it out explicitly (and ideally comment the override to say why it's set, since the field otherwise looks cosmetic).

4. gsa_blob_test.go deletion is still unjustified. Your reply covers TestKeySignerImplementsBlobSigner — and replacing it with the var _ Signer = (*GSASigner)(nil) interface guards is strictly better. But the other three/four tests were dropped, not rewritten black-box, and the package now has no GSA test at all (gsa.go, key.go, key_test.go, signer.go). The code they covered is unchanged and still present:
- TestGSASigningConfigRekorV2RequiresTSA guarded the invariant documented at gsa.go:65-70 ("Setting RekorURL requires TSAURL, since Rekor v2 does not timestamp entries") — a security-relevant config check on the production signing path, added one commit ago in ae757b5.
- TestGSASigningConfigExplicitEndpoints covered gsaSigningConfig service-version selection (service(opts.RekorURL, 2), service(opts.TSAURL, 1)).

gsaSigningConfig is unexported, so black-box testing isn't an option without an export_test.go. Please either restore the file with //nolint:testpackage, or add an export_test.go shim — but don't leave GSA signing config untested.

5. else after return at registry.go:349-359. Lint doesn't flag it (revive's early-return rule is off), but it's unidiomatic and inconsistent with the rest of the file. Invert to if signatureErr != nil { log; } else { return … }, or log-and-fall-through.

6. Two adjacent if evidencePublisher != nil blocks in PublishInstallerIndex — merge them.

Things worth documenting

- The "signature ⟹ evidence" invariant is now load-bearing and correctly enforced (Publish → Verify → SignImage → VerifyImage → promote tag), with the ordering pinned by TestPublishInstallerIndexPromotesOnlyCompleteEvidence. Good. But it only holds for indexes signed by this version and for Talos ≥ 1.13. Installer images already in the registry carry a valid signature and no evidence, and will be served that way indefinitely with no backfill path. The doc currently says "Cache hits are accepted when the cached index has a valid completion signature" without noting that pre-existing images can never acquire evidence. One sentence would close the gap.

Test coverage

Better than the previous revision: signer_test.go's blob tests were carried into key_test.go (no loss there), the interface guards replace the reflection test, and TestEnhanceFromSchematicCapturesConsumedDependencies covers the new capture path including the profile-hash change. Remaining gaps:
- No GSA tests at all (above).
- No test for the overlay path through EnhanceFromSchematicWithDependencies — which is exactly where the duplicate is.
- InstallerEvidenceSupported is tested only at 1.12.9 / 1.13.0; add a pre-release case and it fails today.
- The verifyBundledSignature predicate tightening in internal/image/verify/verify.go still has no test, despite affecting verification of every upstream image the factory pulls.

Verdict

The two structural concerns from last round — provenance materials not being tied to the build, and evidence verification on the hot path — are both properly fixed, and the fixes are the right shape rather than workarounds. What's left is one likely-unintended pre-release gate bug (#1), one duplicate-material bug (#2), the undocumented cache-key change (#3), and the still-unexplained GSA test deletion (#4). Everything else is polish.

smira (approved):
🆒

PR #523, opened 2026-08-12
test: fix two flaky checks in the enterprise integration job

rsmitty:
This PR fixes two integration tests that fail for reasons of their own rather than anything
in the product. https://github.com/siderolabs/image-factory/pull/520 fixed the publish ordering behind one of them, but this part survived.

testScannerDBRotation runs a second factory in the same process, and every factory gets a
fresh cache signing key. Sharing one SPDX cache repository meant both built the same tag,
whoever tagged last won, and the loser's read-back could not verify the winner's signature.
It now gets its own repository, the way it already keeps its metrics namespace and scanner
database to itself, and setupEnterprise leaves a caller-chosen repository alone so the
override sticks.

TamperedTokenRejected flipped the last character of the download token. base64url packs the
64-byte ES256 signature into 86 characters, so that character carries four significant bits
and several values decode to the same bytes, leaving the token valid and the request a 200.
It now flips a bit in the decoded signature instead.

### DIFF
--- a/internal/integration/auth_test.go
+++ b/internal/integration/auth_test.go
@@ -9,12 +9,14 @@ package integration_test
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


smira (approved):

shanduur (approved):

rsmitty:
Amended to also fix a flake in the tampered token test.

PR #524, opened 2026-08-12
fix: record the status the client actually received

rsmitty:
The request log and the audit record inferred the status from the error a handler
returned, so anything that answered for itself was recorded as a 200: redirects,
error pages, and every proxied registry response, including an upstream 404 or 500.
Wrap the response writer with felixge/httpsnoop and read the status off it instead.

Two more pieces land without callers, both for browser login. enterrors.RespondedTag
drops an already-answered error from error to warning level. The wrapper also pins the
Cache-Control the auth provider chose, on every path that commits the header, so a
later handler cannot weaken a no-store.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/anchore/syft v1.50.0
 	github.com/blang/semver/v4 v4.0.0
 	github.com/coreos/go-oidc/v3 v3.20.0
+	github.com/felixge/httpsnoop v1.1.0
 	github.com/fsnotify/fsnotify v1.10.1
 	github.com/go-jose/go-jose/v4 v4.1.4
 	github.com/google/go-containerregistry v0.21.9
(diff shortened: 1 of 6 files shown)


internal/frontend/http/statuswriter.go
```diff
@@ -0,0 +1,75 @@
...
+package http
+
+import (
+	"io"
+	"net/http"
+)
+
+// statusWriter records the status a handler wrote, so the request log and the audit record
+// report what the client received rather than inferring it from the returned error.
+type statusWriter struct {
```
smira: I don't think we should try to reimplement this, but rather use https://github.com/felixge/httpsnoop

This provides proper support for edge cases

rsmitty: good shout, thx. reworked and amended to use httpsnoop for this.

shanduur (approved):

PR #530, opened 2026-08-14
chore: gate the heavy CI jobs on lint and unit tests

rsmitty:
This PR makes the five integration jobs and chart-e2e wait for lint and unit-tests instead of starting alongside them. The verdict lands early either way, but today the six large jobs keep going for another ten minutes on a commit we already know is broken, and holding the pool that long delays unrelated jobs. Omni stages its jobs this way, so this is the same needsOverride it uses. Green runs pay about ninety seconds for it.


smira:
I'm kind of neutral on this - this way broken unit-tests don't let me see other failures? Same with lint - I might do a hacky change, which is not clean on lint, but I want integration test results which are harder to run in local mode.

I think the real answer is to never submit PRs with broken lint/unit-tests which are easy to run locally.

Btw, this is definitely not `feat:`, this is `chore:`

frezbo (approved):

rsmitty:
Good call on the commit msg. Will fix and amend.

I see your reasoning on the rest. Maybe we let @shanduur make the call? Do we want to continue running the fat tests even if the easy ones fail? FWIW, I believe Talos does more of what we have today while Omni leans more into the separation proposed in this PR.

My reasoning here was that we're keeping keeping a whole boat load of cores busy for a PR that isn't even mergeable. But you guys face this kind of problem more than I do, so I'm okay if we say this isn't necessary.

smira:
I don't exactly love Omni's approach for the reasons I stated above - if I _want_ to run integration tests, I have to fix unit-tests/lint, even though they are totally irrelevant for me.

If I messed up IF's lint and I want it to be clean - notice the failure in the CI, quickly re-push the PR, all builds are canceled. 

So it's certainly a matter of preference more.

shanduur (approved):

shanduur:
I agree with the approach - failing early here seems worth the ~90-second cost on green runs, especially if it frees up the pool sooner when lint or unit tests fail.

Especially that we don't have as granular control over what will be run as in Talos.

PR #532, opened 2026-08-17
feat: enforce public API with OpenAPI

### LINKED ISSUE #477: OpenAPI spec
It would be nice to have some consumable spec for the image factory API. Something that could be used to automate tasks or understand what's available.

I think a published spec would be the correct approach, but open to ideas on alternatives.

This would allow more people to use various tools like postman to understand the API and validate tools when the api changes.

shanduur:
Embed and serve a canonical OpenAPI 3.1 contract, route and validate documented requests through kin-openapi, and preserve existing authentication, handler-owned status behavior, request-body compatibility, operational routes, greedy OCI repository paths, and the current application configuration shape.

Closes #477 
Closes #201 

This needs to be followed by #537 


smira (approved):
I think it's very good,  I guess we need to rebase & update it with API tokens.

I wonder if we could somehow assert that no new methods were added without putting them to the OpenAPI spec?

PR #538, opened 2026-08-19
feat: sign installer profile when unified secureboot supported

### LINKED ISSUE #536: unify installer/installer-secureboot outputs for recent Talos
tl;dr make `installer` same as `installer-secureboot` (always sign the installers) if Image Factory has SecureBoot feature enabled.

If IF SecureBoot feature is disabled, `installer` should work and `installer-secureboot` should fail.

But if SecureBoot is enabled, for installers starting with Talos version which ships UKIs in the installer (quirks.SupportsUnifiedInstaller, >=1.10), we can make both profiles generate signed output (only for installers).

Benefits:

* sharing the cache
* supporting the case when someone disabled/enabled SecureBoot, the TPM encryption of the disk should still work

shanduur:
Unified installer contains UKI regardless of secureboot flag, so
standard and secureboot installer profiles built same binary. Without
this, both variants get built/cached separately despite identical
output.

Closes #536


internal/integration/download_test.go
```diff
@@ -428,6 +428,10 @@ func defaultKernelArgsSuffix(talosVersion string) string {
 	return suffix
 }
 
+func secureBootKernelArgsSuffix(talosVersion string) string {
+	return defaultKernelArgsSuffix(talosVersion) + " lockdown=confidentiality"
```
smira: now I wonder if it's good or not to sign all installers.

I totally forgot we have different cmdline for them

smira: it might be good to sign the installer, but I don't think we should enable `lockdown=confidentiality` if it's not SecureBoot (?). it'd be a breaking change in the current shape

internal/profile/profile.go
```diff
@@ -592,6 +604,14 @@ func EnhanceFromSchematic(
 	case profile.OutKindInitramfs, profile.OutKindKernel:
 	// installer and UKI support extra kernel args if either secure boot is enabled or the version supports unified installer
 	case profile.OutKindInstaller, profile.OutKindUKI:
+		if autoSignInstaller {
+			// Talos imager uses SecureBoot both to sign UKIs and to add its lockdown
+			// argument. Keep the established standard-installer cmdline while still
+			// signing; explicit schematic arguments are applied afterwards and can
+			// opt back into lockdown.
+			prof.Customization.ExtraKernelArgs = append(prof.Customization.ExtraKernelArgs, "-lockdown")
```
smira: I believe this is a wrong level to solve this problem - it would clear any user-provided lockdown args as well. We need to adjust default args here.

We need to address this on imager/Talos side imho.

shanduur:
~~I think we need `v1.14.0-rc.2` for this.~~

Already in `main`, no need to pull it here.

frezbo (approved):

PR #539, opened 2026-08-19
feat(enterprise): theme and translate Auth0 logout/login-error pages

rsmitty:
This PR moves the /logout and login-error pages out of inline HTML in browser.go into a new pkg/enterprise/pages package with real templates and i18n, matching the rest of the wizard's look and locales.

Pulling it into its own package also means any future auth plugin can reuse these pages instead of rolling its own.


rsmitty:
@shanduur @smira - not sure who did the original translations for the text we use, but PTAL a what AI generated for these few new translations.

smira (approved):
Russian part looks good ;)

shanduur (approved):
🇵🇱 looks fine

PR #545, opened 2026-08-24
feat(enterprise): add Auth0 Management API client for node tokens

rsmitty:
This PR adds a Management API client for creating, listing, and deleting per-org Auth0 M2M "node token" apps, using the go-auth0 SDK throughout. CreateNodeClient now also requires MachineScope to be configured, so we never accidentally mint a token with full access to everything.


enterprise/auth/auth0/management.go
```diff
@@ -0,0 +1,228 @@
...
+	// managementTokenFetchTimeout bounds the client-credentials token source's HTTP client,
+	// captured once at construction, since it isn't covered by a later per-request timeout.
+	managementTokenFetchTimeout = 10 * time.Second
+
+	managedByTag = "image-factory"
+)
+
+// managementSDKTestOptions lets export_test.go disable the SDK's default 5xx/429 retry
+// backoff, so error-case tests don't wait through it.
+var managementSDKTestOptions []auth0option.RequestOption
```
smira: can we pass it explicitly into the `newSomething` call from the test instead of using global bariables

enterprise/auth/auth0/management.go
```diff
@@ -0,0 +1,228 @@
...
+
+func newManagementSDK(ctx context.Context, tenantURL *url.URL, cfg Config) (*auth0mgmtclient.Management, error) {
+	tokenCtx := context.WithValue(ctx, oauth2.HTTPClient, &http.Client{Timeout: managementTokenFetchTimeout})
```
smira, line 49: using http.Client timeout is an anti-pattern, and instead context should have a timeout as needed? where does it come from? does the client have no proper context handling?

enterprise/auth/auth0/management.go
```diff
@@ -0,0 +1,228 @@
...
+
+	opts := append([]auth0option.RequestOption{
+		auth0option.WithBaseURL(tenantURL.JoinPath("/api/v2").String()),
+		auth0option.WithClientCredentials(tokenCtx, cfg.ClientID, cfg.ClientSecret),
+		auth0option.WithHTTPClient(&http.Client{Timeout: managementRequestTimeout}),
```
smira: same here, I would rather use an HTTP client with idle connection pooling, and one client for all of it

enterprise/auth/auth0/management.go
```diff
@@ -0,0 +1,228 @@
...
+// failure modes via errors.Is, e.g. to map both to an HTTP 404.
+var (
+	ErrInvalidNodeClientID = errors.New("auth0: invalid node client id")
+	ErrNodeClientNotFound  = errors.New("auth0: node client not found")
+)
+
+// ErrMachineScopeRequired is returned by CreateNodeClient when MachineScope is unconfigured.
+var ErrMachineScopeRequired = errors.New("auth0: machineScope must be configured to create node clients")
+
+var validNodeClientIDPattern = regexp.MustCompile(`^[A-Za-z0-9]+$`)
```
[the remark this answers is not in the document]
rsmitty: this doesn't seem to be true. clientIDs returned from auth0 match the original regex. there are no dashes or underscores.

enterprise/auth/auth0/management.go
rsmitty: yep, fixed in amended commit.

enterprise/auth/auth0/management.go
rsmitty: you're right, it's not required here. fixed in amended commit.

enterprise/auth/auth0/management.go
rsmitty: so my understanding here is that there's basically some buggy behavior with the way oauth library does token refreshes. most of the oauth client calls repect the ctx, but token refresh doesn't and this is the only way to protect against a hung refresh.

some info from claude explaining:

```
  Every real API call (Clients.Create, Clients.Get, Clients.List, etc.) does respect your per-call ctx — the SDK builds that outbound HTTP request with
  http.NewRequestWithContext(ctx, ...), so the actual "send this request to Auth0, wait for a response" step is fully bounded by whatever timeout you
  passed in. That's the normal, well-behaved path, and it's most of what happens on every call.

  But before that request can go out, the transport needs a valid Bearer token to attach. That's oauth2.Transport.RoundTrip, and here's the exact code
  again:

  func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
      ...
      token, err := t.Source.Token()   // <-- separate, hidden network call — no context passed
      ...
      req2 := req.Clone(req.Context()) // <-- THIS carries your ctx, used for the real request below
      ...
      return t.base().RoundTrip(req2)  // <-- the real API call, properly bounded by ctx
  }

  req (which does carry your context) is right there in scope — but t.Source.Token() is called with zero arguments, because that's the TokenSource
  interface's signature. So when a token needs fetching or refreshing — a separate POST to /oauth/token, not the request you actually asked for — that
  specific sub-call has no path for your context to reach it at all. It's invisible plumbing that happens before your real call, not your real call
  itself.

  In practice this only bites intermittently: once a token is cached (default ~24h lifetime), most calls skip the fetch step entirely and go straight
  through the properly-context-bounded path. It's only the occasional refresh that's exposed — which is exactly why it needs its own independent bound
  rather than relying on whichever caller happened to trigger that particular refresh.
```

rsmitty: comically old github issue describing this: https://github.com/golang/oauth2/issues/262

smira: thanks for explaining it 🤦‍♂️

smira (approved):

PR #555, opened 2026-08-27
feat: accept download token on /pxe/ and forward it

shanduur:
iPXE cannot send request headers, so a script fetched with a download
token previously had to fall back to Basic userinfo in its asset URLs,
putting a non-expiring credential in the boot script. Forwarding the
same token as ?token= keeps the script's lifetime bound to the token;
nothing is minted, so re-fetching cannot extend it.

- tokenAcceptedOn now covers /pxe/, GET only (HEAD is unregistered)
- Cache-Control: no-store on every authenticated /pxe/ response, which
  also stops the Basic-auth script body from being cached

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -170,7 +170,8 @@ The lifetime can be requested with the `ttl` query parameter as a Go duration, e
 It must fall within `authentication.downloadTokenTTL.min` and `authentication.downloadTokenTTL.max`
 (`30s` and `8h` by default); anything else is rejected with `400`.
 
-The token is appended to an image URL as `?token=<access_token>` and is accepted only on `GET` and `HEAD` under `/image/`.
+The token is appended to an image or PXE URL as `?token=<access_token>` and is accepted only on `GET` and `HEAD` under `/image/`, and on `GET` under `/pxe/`, the only method that route registers.
+A `/pxe/` request forwards it into the asset URLs of the script it returns.
 One token covers every schematic owned by the caller, not just the URL it is used with; see [Authentication](authentication.md#download-tokens).
 
 ### `GET /.well-known/jwks.json`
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -170,7 +170,8 @@ The lifetime can be requested with the `ttl` query parameter as a Go duration, e
 It must fall within `authentication.downloadTokenTTL.min` and `authentication.downloadTokenTTL.max`
 (`30s` and `8h` by default); anything else is rejected with `400`.
 
-The token is appended to an image URL as `?token=<access_token>` and is accepted only on `GET` and `HEAD` under `/image/`.
+The token is appended to an image or PXE URL as `?token=<access_token>` and is accepted only on `GET` and `HEAD` under `/image/`, and on `GET` under `/pxe/`, the only method that route registers.
```
Orzelius, line 173: > the only method that route registers.

Is this needed?

docs/api.md
```diff
@@ -554,13 +555,19 @@ Returns an iPXE script which downloads and boots Talos Linux with the specified
...
-| required | not accepted | denied        | enforced  |
+| Auth     | `?token=` | Machine scope | Ownership |
+| -------- | --------- | ------------- | --------- |
+| required | accepted  | denied        | enforced  |
 
 The script embeds the schematic's kernel command line, which is schematic-derived customization, so a machine-scoped token is denied here for the same reason it cannot read a schematic definition.
-Download tokens are not accepted either.
-iPXE cannot send request headers, so an authenticated boot puts the credential in the URL: `https://<user>:<password>@pxe.talos.dev/pxe/...`, which the client encodes into `Authorization: Basic`.
+
+iPXE cannot send request headers, so the credential goes in the URL, and whichever one authenticated the request is carried into the asset URLs of the returned script:
```
Orzelius, line 564: idk, maybe this is a bit cleaner

```suggestion
iPXE cannot send request headers, so the credential goes in the URL and is carried into the asset URLs of the returned script:
```

docs/authentication.md
```diff
@@ -186,8 +186,11 @@ Verification allows a further 30s of clock leeway.
 ### The `?token=` query parameter
 
 - Its value is the `access_token` from `POST /download-token`, and nothing else.
-- It is accepted only on `GET` and `HEAD` under `/image/`.
-  It does not work on `/pxe/`, on the `/v2/` registry, on schematic routes, or on the SBOM, VEX and scan routes: elsewhere the parameter is ignored and the [`Authorization` header](#the-authorization-header) is the only credential.
+- It is accepted only on `GET` and `HEAD` under `/image/`, and on `GET` under `/pxe/`, the only method that route registers.
```
Orzelius, line 189: same here

frezbo (approved):

Orzelius (commented):
sorry I didn't make it in time, but had a few small nits about the docs

PR #558, opened 2026-08-27
feat(enterprise): add self-issued node tokens

rsmitty:
This PR adds the actual node-token backend. Three parts:
- self-issued JWTs, similar to the download tokens, but with longer TTL and different audience
- a per-org index kept in the registry so tokens can be listed and revoked
- the /node-tokens API to drive it all.

Verification is checked inline the same way download token verification already is, so it doesn't care which auth provider is configured. Storage defaults to the same Sidero-hosted registry schematics already use, so self-hosted deployments don't need anything new to run it. Node token creation isn't Auth0-only. Any configured auth provider will enable it. Although, if we ever ship a provider that doesn't have any mapping to org IDs, we'll need to revisit.

### DIFF
--- a/enterprise/nodetoken/frontend.go
+++ b/enterprise/nodetoken/frontend.go
@@ -1,0 +1,206 @@
+// Copyright (c) 2026 Sidero Labs, Inc.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+
+//go:build enterprise
+
+// Package nodetoken provides HTTP handlers and backing storage for self-serve, self-issued
+// node token management.
+package nodetoken
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/julienschmidt/httprouter"
+)
+
+// maxCreateBodyBytes bounds the /node-tokens POST body, which only ever needs to carry a name.
+const maxCreateBodyBytes = 1 << 12
+
+// NodeTokenManager is the subset of Manager used by the HTTP frontends.
+type NodeTokenManager interface {
+	CreateNodeToken(ctx context.Context, orgID, name string) (id, token string, err error)
+	ListNodeTokens(ctx context.Context, orgID string) ([]Record, error)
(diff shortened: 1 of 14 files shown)


enterprise/nodetoken/frontend.go
```diff
@@ -0,0 +1,206 @@
...
+
+// Methods implements enterprise.FrontendPlugin.
+func (f *RevokeFrontend) Methods() []string {
+	return []string{http.MethodPost}
+}
+
+// Path implements enterprise.FrontendPlugin.
+func (f *RevokeFrontend) Path() string {
+	return "/node-tokens/:id/revoke"
+}
```
Slessi, line 182: This is fine, but as a nitpick could also be `HTTP DELETE` to `/node-tokens/:id`

enterprise/nodetoken/storage.go
```diff
@@ -0,0 +1,355 @@
...
+
+			continue
+		}
+
+		reread, err := s.readIndex(ctx, orgID)
+		if err != nil {
+			lastErr = err
+
+			continue
+		}
```
Slessi, line 221: i think there is a chance that if this reread fails but the push worked, that it might retry and insert the token twice, but maybe thats ok

Slessi (approved):
I think its ok, if the storage solution doesn't work out at least its abstracted behind an interface, should be easy to replace

Slessi (commented):
Maybe we should add methods on the factory client, to consume by Omni? To create node tokens. I guess that will be needed?

enterprise/nodetoken/storage.go
rsmitty: fixed in amended commit.

enterprise/nodetoken/frontend.go
rsmitty: Chose to leave this as-is for now. I think that's valid but I think it requires changing some of the shared routing infra to support DELETEs, not just the node token stuff.

rsmitty:
> Maybe we should add methods on the factory client, to consume by Omni? To create node tokens. I guess that will be needed?

yes, for sure. I'll chase this PR with that.

PR #563, opened 2026-08-28
feat(enterprise): add a node tokens page

rsmitty:
This PR adds the /ui/node-tokens page: list/create/revoke, with a one-time secret modal showing a ready-to-use registry auth config. Two other pieces ride along:
- a small header nav with a logout link, since node tokens needed somewhere to live
- a wizard/node-tokens polish pass: tighter spacing, selectable cards, a sticky footer, scroll fixes, custom modals instead of native confirm(), and missing dark-mode text colors


rsmitty:
little walkthrough of the new experience:

[Screencast From 2026-08-28 10-17-52.webm](https://github.com/user-attachments/assets/2336f86a-c709-48c3-a01d-363524ce9049)

internal/frontend/http/browserlogin.go
```diff
@@ -23,3 +23,11 @@ func (f *Frontend) registerBrowserLogin(registerPublicRoute func(func(string, ht
...
 	registerPublicRoute(f.router.GET, blp.CallbackPath(), blp.CallbackHandler())
 }
+
+// browserLoginEnabled reports whether the auth provider serves the interactive login flow,
+// gating whether pages show the nav bar's logout/node-tokens links.
+func (f *Frontend) browserLoginEnabled() bool {
+	blp, ok := f.options.AuthProvider.(enterprise.BrowserLoginProvider)
+
+	return ok && blp.BrowserLoginEnabled()
+}
```
Slessi: I get the idea but i think its more clear if instead of `browserLoginEnabled` its something like `factoryRequiresAuth` - sounds a bit closer to the requirement

internal/frontend/http/templates/wizard-cmdline.html
```diff
@@ -112,7 +112,7 @@
     </div>
     {{ end }}
 
-    <div class="flex gap-4">
+    <div class="sticky bottom-0 z-10 -mx-6 mt-6 flex gap-3 border-t border-gray-200 bg-white/90 px-6 py-4 backdrop-blur dark:border-gray-800 dark:bg-gray-900/90">
```
Slessi: maybe this bottom bar should be a common.html template?

Slessi (approved):

internal/frontend/http/templates/wizard-cmdline.html
rsmitty: fixed this in amended commit.

internal/frontend/http/browserlogin.go
rsmitty: this actually wound up being kind of a bug. reworked in amended commit such that any auth provider configured allows for node-tokens to be used. then we additionally determine if logout button is shown based on whether that auth provider implements the logout routes (only auth0 for now).

shanduur (approved):

shanduur (approved):

PR #566, opened 2026-09-01
feat(enterprise): redesign API token authorization

shanduur:
Download and node tokens were the same ES256 issuer differing only
by audience, key and config block, and Scope was a fixed enum that
couldn't express new resource types or finer sub-actions. One
credential now carries a scope claim checked against scopeRoutes,
the single source of truth for what each scope reaches; Auth0
machine scopes and CLI-issued admin tokens read the same table.

- scopes: schematic, token (attenuating, never grantable, caps
  minting blast radius), admin (CLI-only via `admin-token`
  subcommand, never stored, never mintable over HTTP, may name a
  cross-tenant `subject`)
- TokenTTLOptions collapses per-scope bounds into
  Stored/Ephemeral/Bootstrap; TokenOptions.KeyPath becomes KeyPaths
  for rotation without downtime
- /ui/tokens replaces free-form scope checkboxes with four
  code-owned actor profiles (talos/automation/operator/admin) so a
  browser can't hand-assemble sensitive combos like token:issue
- pkg/client DownloadToken becomes TokenCreate with the download
  scope
- integration tests cover minting, per-scope route access, listing,
  revocation, and the grant/admin restrictions

> [!WARNING]
> **BREAKING CHANGE**: node tokens issued by v1.6.0 are rejected; re-issue
from /ui/tokens using scopes: ["pull"]. /node-tokens routes are
removed. authentication.downloadTokenKeyPath,
authentication.downloadTokenTTL and enterprise.nodeTokens move to
authentication.tokens (single keyPath now signs every scope; pick
one of the old download/node keys), rejected at startup if still
set.

---

image
image
image
image


docs/api.md
```diff
@@ -139,17 +141,95 @@ Supported report formats:
...
+```json
+{"tokens": [{"id": "0d1c...", "name": "rack-3", "scopes": ["pull"], "created_at": "...", "expires_at": "..."}]}
+```
+
+Download-scoped tokens never appear: they are not recorded, because they expire long before
+anyone could want them back.
+Only recorded tokens count against `authentication.tokens.maxPerOrg`, and a `POST` that would
+exceed it is `409`.
+
+### `POST /tokens/:id/revoke`
```
smira, line 213: not that it's a big issue, but I wonder why not `DELETE /tokens/:id` ?

docs/api.md
```diff
@@ -139,17 +141,95 @@ Supported report formats:
...
+This route stays for clients built before that changed.
 
 Access:
 
-| Auth     | `?token=`    | Machine scope | Ownership      |
-| -------- | ------------ | ------------- | -------------- |
-| required | not accepted | denied        | not applicable |
+| Auth | Scopes | Ownership |
+| -------- | ------ | --------- |
+| required | `token` | not applicable |
```
smira: do we still need download tokens? or should we just use tokens with download/pull scope?

docs/api.md
shanduur: Because semantically in REST revocation is usually a state transition, not resource deletion. This is IMHO more standard approach.

rsmitty: iirc, I ran into this same question from edward and it was just a case of us not handling any DELETEs in the router to this point and not wanting to thread it through. but it's certainly doable.

smira (commented):
- add "stored: true/false" status to JWT (whether the token is stored, and revokable)
- on verification path, after validating JWT, check if it's stored, and it if is -> check token storage
- on token generation, if TTL < T1 => only non-revokable, if TTL >= T2 => only revokeable
- merge download & API tokens into "API tokens", allow them as query args (in addition to headers)
- IF UI can only generate revokable tokens (TTL >= T1)

smira:
* probably UI should allow you to set the TTL? (and we should show the remaining TTL in the list)

utkuozdemir:
Nit: commit and PR descriptions say
> Enforced on both /tokens and the /download-token alias.

But the `/download-token` is gone, so that info is stale.

hack/release.toml
```diff
@@ -19,6 +19,60 @@ Unified installers embed a UKI regardless of the SecureBoot flag, so `installer`
 and share cached assets instead of being built and stored twice.
 
 Talos versions before 1.14.0 are unaffected: signing an installer there also turns on `lockdown=confidentiality`, so the two variants stay distinct.
+"""
+
+    [notes.api_tokens]
+        title = "API Tokens"
+        description = """\
+Download tokens and node tokens are now one credential, an API token, and its scopes decide what it may reach.
+There are four. `download` authenticates image downloads and PXE scripts, `pull` authenticates the OCI registry a Talos node fetches installers from, `schematic` creates and reads schematics, and `token` mints and revokes tokens.
```
utkuozdemir: Five now, the second commit added `admin` and the next note describes it. Same for "all four" on the next line.

hack/release.toml
```diff
@@ -19,6 +19,60 @@ Unified installers embed a UKI regardless of the SecureBoot flag, so `installer`
...
+One key signs all four.
+
+`token` is the only scope that hands out authority, so it is limited in two ways.
+A token may not grant a scope it does not hold, and may not grant `token` at all.
+A stolen minting credential can spend its own lifetime, but it cannot mint a replacement that outlives revoking it.
+
+The Auth0 machine scope now reads the same route table as `pull`, so widening one widens the other.
+
+Each token also carries a `stored` claim saying whether the factory keeps a record of it, decided once when it is minted rather than inferred from its scopes on every request.
+A stored token is in the per-org index, which is what lists it, what keeps it valid and what a revoke removes. An unstored one is a signed string and nothing else. It needs no name, costs no registry write, and is the only kind accepted from a `?token=` query parameter, on any scope.
```
utkuozdemir: Not on any scope: `URLSafe` refuses `token` and `admin` from the query string regardless of `stored`, and `docs/authentication.md` says so. This sentence should match.

internal/apitoken/apitoken.go
```diff
@@ -0,0 +1,434 @@
...
+	return ttl, nil
+}
+
+// storageRule names the storage bound that narrowed the window, so a rejection says which rule
+// fired. It is empty when no storage bound applied and the scopes' own window is the whole story,
+// which would otherwise blame UnstoredMax for a ceiling it did not set.
+func storageRule(scopes []Scope, stored bool) string {
+	switch {
+	case stored:
+		return "the minimum lifetime of a stored token"
```
utkuozdemir: Even when the requested ttl is above the max, this says "minimum lifetime".

enterprise/tokens/storage.go
```diff
@@ -0,0 +1,360 @@
...
+		}
+
+		stored, readErr := s.readRecord(ctx, s.repository.Tag(tag))
+		if readErr != nil {
+			return nil, fmt.Errorf("tokens: failed to read record tag %q for org %q: %w", tag, orgID, readErr)
+		}
+
+		// Verify the embedded org ID as well as the hashed tag prefix so even a theoretical
+		// prefix collision cannot expose another organization's record.
+		if stored.OrgID != orgID || stored.Revoked {
```
utkuozdemir: Expired tokens are not filtered here, so they stay in the listing and keep counting against `maxPerOrg` until revoked manually. Or am I missing something?

Reproducer, save it beside `enterprise/tokens/manager_test.go` and run:

```console
go test -tags=enterprise ./enterprise/tokens/ -run TestReviewExpiredTokenStaysListed
```

```go
func TestReviewExpiredTokenStaysListed(t *testing.T) {
	mgr := tokens.NewManager(newTestIssuer(t), newTestStorage(t, time.Minute))

	_, _, err := mgr.Create(t.Context(), "org_a", "short", downloadScope, true, time.Second)
	require.NoError(t, err)

	time.Sleep(1500 * time.Millisecond)

	records, err := mgr.List(t.Context(), "org_a")
	require.NoError(t, err)
	require.Empty(t, records, "expired at %s", records[0].ExpiresAt.Format(time.RFC3339))
}
```

At this head it fails with `Should be empty, but was [{... short [download]}]`.

internal/apitoken/scope.go
```diff
@@ -0,0 +1,185 @@
...
+	ScopePull: func(method, path string) bool {
+		return readOnly(method) &&
+			(strings.HasPrefix(path, "/image/") || path == "/v2" || strings.HasPrefix(path, "/v2/"))
+	},
+	ScopeSchematic: func(method, path string) bool {
+		if method == http.MethodPost {
+			return path == "/schematics"
+		}
+
+		return readOnly(method) && strings.HasPrefix(path, "/schematics/")
```
utkuozdemir, line 46: No scope reaches `/spdx/`, `/vex/` and `/scans/`, so a token cannot fetch the security reports. Omni fetches them and proxies them to its UI, so it needs a scope that covers them. Should they go under `schematic`? They are per-schematic reads with the same ownership check.

enterprise/tokens/storage.go
```diff
@@ -0,0 +1,360 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	delete(s.cache, orgID)
+}
+
+func (s *Storage) readRecords(ctx context.Context, orgID string) ([]Record, error) {
+	options := append(slices.Clone(s.remoteOpts), remote.WithContext(ctx))
+
+	tags, err := remote.List(s.repository, options...)
```
utkuozdemir: If I understand correctly, this lists every tag in the repository for all orgs and then filters for one org, and it runs on every cache refresh of a stored token. Since the tag name of a token is deterministic, can we make `Valid` read the exact `tokenTag(orgID, jti)` instead, as an optimization? Then the listing is only needed for `GET /tokens` and the cap check. Also, a freshly created token would work on all replicas immediately, without waiting for the cache to expire.

internal/apitoken/apitoken.go
```diff
@@ -0,0 +1,434 @@
...
+	bounds := TTL{Min: math.MaxInt64, Max: math.MaxInt64, Default: math.MaxInt64}
+
+	for _, scope := range scopes {
+		scopeTTL, ok := i.ttl[scope]
+		if !ok {
+			return 0, fmt.Errorf("apitoken: no TTL configured for scope %q", scope)
+		}
+
+		bounds.Min = min(bounds.Min, scopeTTL.Min)
+		bounds.Max = min(bounds.Max, scopeTTL.Max)
```
utkuozdemir: Let me make sure I get the Omni side right. Omni gets a long-lived token passed to it by omni-controller, scoped to its org, with these scopes, right?

- `schematic`, to create and read schematics.
- `token`, to create download tokens for media links and pull tokens for machine configs, via `POST /tokens`.
- `download` and `pull` themselves, since `CanGrant` only allows granting what the caller holds.

If so, with the defaults, this line breaks that token:

1. `download.max` is 8h, and this takes the smallest max across the scopes, so the Omni token cannot live longer than 8h. Without a ttl it gets 1h.
2. Asking for 30 days fails with `requested TTL out of range: 720h0m0s is outside [1h0m0s, 8h0m0s]`.
3. Omni cannot renew itself either, since a token is not allowed to create another token carrying `token`, so it cannot create its own replacement. The only way out would be omni-controller re-injecting a new token into Omni every few hours, and a self-hosted Omni has no way at all.

At the same time, `download.max` does not limit real download links, because those are unstored and capped by `unstoredMax` anyway. So, wondering if the lifetime should follow only stored/unstored, and the per-scope bounds go away. That would also drop the ttl table from the config.

@rsmitty omni-controller will inject these tokens into Omni as config and rotate them, which means an Omni restart on each rotation, right? Let's clarify what the TTL of Omni's token will be, and at which threshold omni-controller renews it.
 Actually, if omni-controller mounts the token as a file (a k8s secret or similar), Omni can watch that file and reload it without a restart.

smira: >  So, wondering if the lifetime should follow only stored/unstored, and the per-scope bounds go away. That would also drop the ttl table from the config.

I agree, we should not limit the TTL for stored tokens aggressively, I think 1y is safe

smira:
Actors accessing the Image Factory:

* Talos itself, or someone going to install/boot Talos (token-based auth):
   - downloading images like ISO, disk image, PXE, ....
   - pulling `installer` image
* Cluster management platforms (e.g. Omni, Terraform Provider) (token-based auth):
   - create schematics, read schematics
   - downloading images, PXE, ...
   - creating lower tier tokens (tighter scope)
   - access security reports
 * On-prem/air-gapped deployments of Image Factory itself (token-based auth):
   - pulling via Image Factory source images: `image-factory` itself, and Talos source images `imager`, ...
 * Operator (interactive auth, Auth0):
   - create schematics, read schematics
   - downloading images, ...
   - create any tokens
   - access security reports

shanduur:
@coderabbitai review

cmd/image-factory/cmd/options.go
```diff
@@ -679,19 +667,54 @@ type AuthenticationOptions struct { //nolint:govet // keeping order for semantic
...
+	// A token minted with "stored": false is not recorded, so it cannot be listed or revoked and does not count against MaxPerOrg.
+	Storage OCIRepositoryOptions `koanf:"storage"`
+
+	// TTL bounds token lifetimes by whether they are stored, plus the CLI-only bootstrap policy.
+	TTL TokenTTLOptions `koanf:"ttl"`
+
+	// VerificationCacheRefreshInterval controls how often the backing registry clients are rebuilt
+	// so refreshed credentials are picked up.
+	//
+	// The legacy name is retained for configuration compatibility.
```
smira: do we need to maintain any legacy?

cmd/image-factory/flags.go
```diff
@@ -19,26 +22,73 @@ var (
...
+		return fmt.Errorf("unknown command %q; run image-factory --help for the commands this build has", extra[0])
+	}
+
+	return nil
+}
+
+// removedConfigKeys maps a config key that no longer exists to the key that replaced it.
+// Keys are lowercased, since koanf lowercases what it reads from the environment provider.
+var removedConfigKeys = map[string]string{
+	"authentication.downloadtokenkeypath": "authentication.tokens.keyPaths",
```
smira: should we just update the configs and have no compat? we don't have it deployed in prod, less code to carry forward

docs/api.md
```diff
@@ -6,23 +6,24 @@ It intentionally excludes the HTML UI and its supporting routes and static asset
...
-| **`?token=`**     | Whether a [download token](authentication.md#the-token-query-parameter) in the query string is accepted in place of the header.                                                                   |
-| **Machine scope** | Whether a token carrying `authentication.auth0.machineScope` may call the endpoint. `denied` is `403`, even though the token itself is valid.                                                     |
-| **Ownership**     | Whether the schematic's `owner` must match the caller identity, where a mismatch is `403`; an unauthenticated caller gets `401` whether or not the schematic exists, so existence is not leaked.  |
+| Field | Values |
+| --- | --- |
+| **Auth** | `public` takes no credential, `required` takes an [`Authorization` header](authentication.md#the-authorization-header) for the configured provider, and a missing or invalid credential is `401`. |
+| **Scopes** | Which [API token](authentication.md#api-tokens) capabilities may call the endpoint, or `none` when only a full credential can; an Auth0 machine credential receives exactly `image:read`. |
```
smira: what is Auth0 machine credential?

internal/frontend/http/pxe.go
```diff
@@ -78,7 +78,7 @@ func (f *Frontend) handlePXE(ctx context.Context, w http.ResponseWriter, r *http
 	}
 
 	// Image asset URLs carry a credential so iPXE can authenticate when fetching kernel/initramfs:
-	// the download token this request was authenticated with, or the Basic credentials it arrived
+	// the image-read token this request was authenticated with, or the Basic credentials it arrived
```
utkuozdemir:
```suggestion
	// the image:read token this request was authenticated with, or the Basic credentials it arrived
```
or is it intentional?

docs/authentication.md
```diff
@@ -155,61 +160,272 @@ Reading a schematic, or any artifact derived from one, requires the `owner` to m
...
+
+1. Append the new key or certificate after the current active key and roll out that verification set everywhere.
+2. Move the new private key to the first position, keep the old key later in the list, and roll out again.
+3. Remove the old verification key only after every token it signed has expired, or when intentionally invalidating those tokens.
+
+Prepending a new key in a single rolling deployment is unsafe: updated replicas can mint tokens that replicas still running the previous configuration cannot verify.
+
+Verification allows 30s of clock leeway.
+
+### Upgrading from separate download and node tokens
```
smira: we don't have anyone running old versions, so I'd drop it

docs/api.md
shanduur: Those are machine-to-machine Auth0 tokens (this is part of the behavior we have since @rsmitty implemented it first). Should we drop it all-together?

smira: I would guess we don't need them anymore, API tokens replace it all.

Machine-to-machine tokens have limited lifetime, so they provide no value really over API tokens from my understanding.

docs/authentication.md
shanduur: Fixed.

docs/api.md
shanduur: Removed.

cmd/image-factory/flags.go
shanduur: Fixed.

cmd/image-factory/cmd/options.go
shanduur: Removed.

internal/frontend/http/pxe.go
shanduur: Thanks, missed that!

enterprise/tokens/manager.go
```diff
@@ -0,0 +1,119 @@
...
+		logger.Debug("credential is not an API token", zap.Error(err))
+
+		return apitoken.Claims{}, false
+	}
+
+	if !claims.Stored {
+		return claims, true
+	}
+
+	valid, err := m.storage.Valid(ctx, claims.Subject, claims.ID)
```
smira: should we cache the lookup for some _short_ time?

it's a balance of correctness and performance

internal/frontend/http/http.go
```diff
@@ -389,32 +375,43 @@ func (f *Frontend) withAuth(h Handler, requireAuth bool, username *string, state
...
-		if f.options.NodeTokenVerifier != nil && (r.Method == http.MethodGet || r.Method == http.MethodHead) && nodeTokenAcceptedOn(r.URL.Path) {
-			if tokenStr := extractBearerOrBasicToken(r); tokenStr != "" {
-				if orgID, ok := f.options.NodeTokenVerifier.Verify(ctx, tokenStr); ok {
-					*username = orgID
-					ctx = authProvider.ContextWithUsername(ctx, orgID)
-
-					return h(ctx, w, r, p)
+		// API token: the JWT subject becomes the authenticated identity, so ownership is then
+		// enforced normally by schematicFactory.Get(). An ephemeral token is also put on the
+		// context, for handlePXE to forward into the asset URLs it emits.
```
smira, line 380: we probably should let PXE pass any tokens (including stored ones), as if the request with the token reached the IF, the token has already been exposed, so it's the risk that a user accepts

enterprise/tokens/storage.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+
+		return cmp.Compare(a.ID, b.ID)
+	})
+
+	return tokens, nil
+}
+
+func (s *Storage) readRecord(ctx context.Context, ref name.Tag) (storedRecord, error) {
+	desc, err := s.puller.Get(ctx, ref)
```
smira, line 202: as a suggestion - what if we put the org ID to the repository name, so the tag becomes the token ID basically? 

this way at least number of tags per repository will be bounded by the max tokens per org?

smira (approved):

enterprise/tokens/manager.go
shanduur: I added single-flight instead, this will keep the number of requests smaller, and not cause any cache invalidation issues.

PR #591, opened 2026-09-17
feat: add 'On Premises' role to the token UI

### LINKED ISSUE #590: ui: add a token preset for `source:pull` scope
Need to pick a name, but basically someone just accessing source images from the EIF.

smira:
Fixes #590


smira:
image

docs/authentication.md
```diff
@@ -172,6 +172,7 @@ It offers fixed actor profiles whose scope lists are code-owned alongside the sc
 | Actor | Executable scopes | Issuable scopes |
 | --- | --- | --- |
 | Talos | `image:read` | none |
+| On Premises | `source:pull` | none |
```
Slessi, line 175: maybe `upstream:read`?

pkg/client/client.go
```diff
@@ -53,6 +53,8 @@ type Actor string
 const (
 	// ActorTalos can fetch generated images, PXE assets, and installer OCI artifacts.
 	ActorTalos Actor = "talos"
+	// ActorOnPrem can pull source images for mirroring into an on-premises installation.
+	ActorOnPrem Actor = "onprem"
```
Slessi, line 57: also this one `onprem` doesnt really describe "what", it describes "where"

maybe `mirror` instead?

docs/authentication.md
smira: The `source:pull` was a pre-defined scope name (I haven't touched it in this PR). I think it's a good name - `source` is "source images" - it describes "what". `upstream` is unclear in this context (what is upstream for _this_ IF?)

pkg/client/client.go
smira: I'm not positive, given that we have "Talos", "Omni" as actors, `mirror` sounds like an action? The actor is more or less a UI concept only, it is not part of the actual API (scope list is the API).

docs/authentication.md
Slessi: sure 👍

pkg/client/client.go
Slessi: It can be both noun & verb, so like: "The mirror is mirroring" - in this case I meant the noun 😄 

but its fine either way, naming is hard

Slessi (approved):

PR #592, opened 2026-09-18
feat: improve the token UI

### LINKED ISSUE #583: ui: show a loading state on the tokens page
The tokens page renders an empty list until the tokens arrive, which on staging took several seconds. An empty list reads as "no tokens", so the user starts creating one that already exists, or wonders where the tokens went.

Show a loading indicator until the list request returns, and the empty state only after that.

Observed on the staging factory, Image Factory Enterprise v1.7.0.

### LINKED ISSUE #584: ui: give feedback when revoking a token
Clicking revoke and confirming leaves the row in the list with no indication that anything happened. It disappears a while later. In between, the user does not know whether the revoke went through and is tempted to click again.

Disable the row or show a spinner while the request runs, remove it as soon as the request succeeds, and show the error if it fails.

Observed on the staging factory, Image Factory Enterprise v1.7.0.

### LINKED ISSUE #585: ui: list common durations in the token lifetime help
The lifetime field takes a Go duration, and the help text below it already says that days are not a unit and 30 days is 720h. That is right, but the user still has to do the math for anything but 30 days: a one year token is 8760h, and nothing on the page says so.

Add the common values to the help text, e.g. as a short list:

- 1 day: 24h
- 30 days: 720h
- 90 days: 2160h
- 1 year: 8760h

Showing the resulting expiry date next to the field would also help.

### LINKED ISSUE #586: ui: show the scopes behind each token actor
The create token dialog offers the actors Talos, Automation, Operator and Admin, each with a one sentence description. The scopes an actor translates to are only visible after creation, in the permissions column. A user picking Talos for an Omni instance ends up with a token that cannot create schematics, and finds out from the failing requests.

Show the scopes each actor translates to in the dialog, next to the description.

### LINKED ISSUE #587: ui: show the server's error message when creating a token fails
Creating a token with a lifetime above the cap fails with a bare "HTTP 400" in the dialog. The response body has the useful part: `apitoken: requested TTL out of range: 87600h0m0s is outside [1h0m0s, 8760h0m0s] (the lifetime of a stored token)`. Only the network tab shows it.

Show the message from the response in the dialog instead of the status code. The same applies to the other failures of that form, e.g. a missing name or a token count above the per-org cap.

Observed on the staging factory, Image Factory Enterprise v1.7.0.

smira:
Multiple small fixes/improvements, none of them touch the backend.

Fixes #583

Fixes #584

Fixes #585

Fixes #586

Fixes #587


smira:
image
image

smira:
image

internal/frontend/http/templates/tokens.html
```diff
@@ -387,6 +554,70 @@ <h2 class="text-base font-semibold mb-2 text-gray-900 dark:text-white">{{ t .Loc
             createErrorEl.classList.add('hidden');
         }
 
+        // durationUnits maps a Go duration unit to milliseconds, which is what Date works in.
+        var durationUnits = {
+            ns: 1e-6, us: 1e-3, 'µs': 1e-3, 'μs': 1e-3, ms: 1, s: 1000, m: 60000, h: 3600000
```
Slessi, line 559: waw couldve sworn `'µs': 1e-3, 'μs': 1e-3` was a duplicate but `"µ" !== "μ"` 

TIL microSign and greek Mu both exist and are different

internal/frontend/http/templates/tokens.html
```diff
@@ -387,6 +554,70 @@ <h2 class="text-base font-semibold mb-2 text-gray-900 dark:text-white">{{ t .Loc
...
+        };
+
+        // parseGoDuration converts the lifetime field to milliseconds so the dialog can say when
+        // the token would expire. It accepts a sequence of amount+unit pairs, the same shape the
+        // server takes, and returns null for anything else: the preview then stays hidden and the
+        // server remains the one deciding what is valid.
+        function parseGoDuration(text) {
+            var re = /(\d+(?:\.\d+)?)(ns|us|µs|μs|ms|s|m|h)/y;
```
Slessi, line 567: i feel like it might be easier (for everyone) if the frontend had 3 inputs for days, hours, minutes, and JS code parses & sums them together to convert to a go duration, rather than this regexy stuff and a list of durations

(#585 asked for the list though i didnt notice a list anywhere)

smira: #585 is there, it's inside translations basically (as it's help text), or you can see in the screenshots

This is used to show the expiration date (expected) in JS before submitting.

Slessi: Yes, I just meant that changing the UX as below would remove the need for regex parsing for us, and negate possible typos from the user. Also no need for hour based math to figure out 1year or so.

Not a blocker though, just a suggestion

image

Slessi (approved):

internal/frontend/http/templates/tokens.html
smira: yeah, I see your point, still needs some validation though, let's see if this version will be good enough. It's still mostly for devops people, as you need to know how to use the token, so hopefully Go's duration will not be that bad (it's consistent with API input).

utkuozdemir:
Good stuff, closed all of my mini bug reports!

PR #593, opened 2026-09-21
feat: update dependencies

smira:
Update Talos to 1.14.1, sync pkgs and update other Go modules.


majabojarska (approved):

go.mod
```diff
@@ -1,71 +1,71 @@
 module github.com/siderolabs/image-factory
 
-go 1.26.7
+go 1.27.0
```
frezbo, line 3: intentional?

frezbo (approved):

go.mod
smira: not really, comes through some of the dependencies (not us)

